PR #108, opened 2019-01-28
Optimize book keeping for violated PL constraint selection

sparth95:
Updating book keeping for violated PL constraints in SMT Core. Each PLconstraint is registered as a watcher and eagerly report the change in status as the participating variables change.


AleksandarZeljic:
We're abandoning this direction for now. Closing the PR.

PR #159, opened 2019-07-29
Property parser scalar check

ahmed-irfan:
This PR introduces:

* a check that right handside of each constraint is a scalar
* minor changes in the property parser


guykatzz:
LGTM

PR #162, opened 2019-08-07
dump input query

yuvaljacoby:


guykatzz:
Made a few cosmetic changes, please review. I compiled but didn't run, please check that it still works. Then you can merge.

PR #164, opened 2019-08-19
Acas parser overflow fix

guykatzz:


kjulian3:
The "+1" for means and ranges is for the neural network output normalization. There should be one entry for every input variable, and an additional value for all output variables, so (nnet->inputSize)+1 should be correct. There's more documentation about NNet at https://github.com/sisl/NNet.

PR #167, opened 2019-08-22
Dump query

yuvaljacoby:


maraboupy/MarabouCore.cpp
```diff
@@ -103,7 +103,9 @@ void createInputQuery(InputQuery &inputQuery, std::string networkFilePath, std::
     printf( "Property: None\n" );
 }
 
-std::pair<std::map<int, double>, Statistics> solve(InputQuery inputQuery, std::string redirect="", unsigned timeout=0){
+/* The default parameters here are just for readability, you should specify
+ * them in the to make them work*/
```
guykatzz: Sentence is broken

maraboupy/MarabouCore.cpp
```diff
@@ -112,14 +114,14 @@ std::pair<std::map<int, double>, Statistics> solve(InputQuery inputQuery, std::s
     if(redirect.length()>0)
         output=redirectOutputToFile(redirect);
     try{
-        Engine engine;
-        if(!engine.processInputQuery(inputQuery)) return std::make_pair(ret, *(engine.getStatistics()));
+        Engine *engine = new Engine(verbosity);
```
guykatzz: if you allocate dynamically, you have to delete() (in every possible branch of the execution).

maraboupy/MarabouCore.cpp
yuvaljacoby: That is to keep it less then 80 characters.
We don't want that?

yuvaljacoby:
Tried to create a test for maraboupy to make sure no memory leaks.
Had trouble with the compilation, I think I will try to simplify the compilation process (cmake) and then get back to write the test, anyway the python changes here are small but are suppose to save time for new users using the python 


A bit more about the compilation difficulties  I had - 
I needed to extend the MockEngine & IEngine (to have solve method) .
Then tried to change the MarabouCore to work with IEngine and then stuff started to break in MarabouCore and in SmtCore tests

yuvaljacoby:
@guykatzz  Can I merge this?

guykatzz:
Yes

PR #169, opened 2019-08-26
make the DnC work with maraboupy.

jjgold012:


AleksandarZeljic:
We are going through pull requests, is this one still active? If yes, can we provide a description so that we can assign it correctly.

jjgold012:
It still alive. I've changed the name of the pull request.

PR #172, opened 2019-09-02
Minor fix for printing

kjulian3:
This is really two very simple PRs merged into one, so I can split this up if needed.

There's a place in Engine.cpp that doesn't  check _verbosity before printing, which is fixed here. 

Also, maraboupy's function for evaluating a network at a point always prints a line that is not useful because it just tells the user that they called the function. When evaluating many points with maraboupy, this can result in a lot of print statements, so I think it's best to remove the print statement.

### DIFF
--- a/src/engine/Engine.cpp
+++ b/src/engine/Engine.cpp
@@ -210,7 +210,10 @@ bool Engine::solve( unsigned timeoutInSeconds )
                     if ( _tableau->getBasicAssignmentStatus() !=
                          ITableau::BASIC_ASSIGNMENT_JUST_COMPUTED )
                     {
-                        printf( "Before declaring SAT, recomputing...\n" );
+                        if ( _verbosity > 0)
+                        {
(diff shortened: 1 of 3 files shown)


src/engine/Engine.cpp
```diff
@@ -210,7 +210,10 @@ bool Engine::solve( unsigned timeoutInSeconds )
                     if ( _tableau->getBasicAssignmentStatus() !=
                          ITableau::BASIC_ASSIGNMENT_JUST_COMPUTED )
                     {
-                        printf( "Before declaring SAT, recomputing...\n" );
+                        if ( _verbosity > 0)
```
guykatzz: Please add a space: "> 0)" --> "> 0 )"

Apart from that, LGTM - you can go ahead and merge.

kjulian3: Added the space, I'll merge this in.

PR #175, opened 2019-09-12
Cmake

yuvaljacoby:
Tested on Mac and Ubuntu 16.04

The main change is moving to CMake, compilation time is ~half, but it's adding a dependency in CMake which does not resolve automatically, should we do it?  (as far as I understand from CVC4 code they also don't do it)

Added ASan for memory leaks detection, and fixed all the memory leaks that it found (most of them were in tests, but one real in LUFactorization).

Changed the python API to auto download pybind (no need to install prior), the only issue is that they are using the latest python you have installed which might cause conflicts (a user can want to use python3.6 in the API while he has 3.7 installed and it will not work).

There are three flags which I set ON by default:
1. Whether to build python API
2. Whether to run tests with ASan enabled
3.  Where to run tests on every build


guykatzz:
It now compiles on my machine (with Python disabled). We will need to fix that issue, though, before we can merge.

Also, the regression tests compile, but don't run - they need some extra files, found in the regress folder.

On the matter of flags:
Debug mode and cxx mode are separate things. In Debug mode we want to set DEBUG_ON (src/common/Debug.h), to get our assertions to run. In cxx mode, we can disable certain warnings that mess with CxxTests. Cxx mode should include Debug mode.

yuvaljacoby:
Fixed regress and debug flags

Regarding python, managed to reproduce, the problem is that the default python is python32 on a 64 bit computer, and then basically we need to compile marabou for 32 bit (only for the python api).
That is of course possible, but breaks the (relatively)clean build process I was trying to create, currently we compile marabou once into a library and the link to it with two API's (executable and Python). If we will change this it makes the build pretty complicated (cannot compile both 32 and 64 in the same CMakeLists.txt)

I can think of two options to solve this:
1. If python build is on and python is 32 bit, build Marabou for 32bit (and add warning probably)
2. Separate the builds and then we will end up with something pretty similar to what we have today, but in CMake

guykatzz:
Hmm... are we only compiling the library once? Some of the debug flags have to be passed at compilation time, not linking. So, I think we should be compiling:
- Once for the executable and python API, say in release mode
- Once for regression tests, with debug 
- Once for ASAN, with debug flags and the ASAN flags. (if these flags are only at linking, then this can be joined with the previous item).

If we are indeed compiling multiple libraries, why not add another one in 32-bit for the python interface? And add flags to control whether it's in use or not.

yuvaljacoby:
Actually we build two instances for the library:
1. Debug + ASan - used for tests
2. Release - for Regress and Executable
They are not exactly the same since the test lib is using the mock files instead of the real (src/common/mock for example)

What you suggested will not work, because  of how pybind is deciding on the compiler version.
I think I found a better solution, if this is the case (i.e. python build is enabled and there is only python32 on 64bit machine), I am not building the python API.
Instead writing a warning message, and pointing to a script that can do that.
The script is maraboupy/build_marabou_x86.sh, basically he builds Marabou for 32bit and expose the python API.

Please make sure it works on your computer, because those stuff are very sensitive (I did manage to reproduce the problem and it was solved but still...)

guykatzz:
Regressions are running - very good.
I'm also no longer getting the Python error on default.

The Python x32 script fails, though:
---------------
./build_python_x86.sh 
-m32
-- The C compiler identification is GNU 8.2.0
-- The CXX compiler identification is GNU 8.2.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Python: /usr/bin/python3.7 (found version "3.7.0") found components:  Interpreter 
CMake Error at /usr/share/cmake-3.12/Modules/FindBoost.cmake:2063 (message):
  Unable to find the requested Boost libraries.

  Boost version: 1.62.0

  Boost include path: /usr/include

  Could not find the following Boost libraries:

          boost_program_options

  No Boost libraries were found.  You may need to set BOOST_LIBRARYDIR to the
  directory containing Boost libraries or BOOST_ROOT to the location of
  Boost.
Call Stack (most recent call first):
  CMakeLists.txt:108 (find_package)


-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - yes
-- Found Threads: TRUE  
-- Found PythonInterp: /usr/bin/python3.7 (found version "3.7") 
-- Found PythonLibs: python3.7m
-- pybind11 v2.3.0
-- Performing Test HAS_FLTO
-- Performing Test HAS_FLTO - Success
-- LTO enabled
-- Found CxxTest: /cs/labs/guykatz/guykatz/cmake/tools/cxxtest  
-- Configuring incomplete, errors occurred!
See also "/cs/labs/guykatz/guykatz/cmake/maraboupy/build32/CMakeFiles/CMakeOutput.log".
See also "/cs/labs/guykatz/guykatz/cmake/maraboupy/build32/CMakeFiles/CMakeError.log".
make: *** No targets specified and no makefile found.  Stop.


----------------

Also, there are some issues with the CxxTests. 

1. Currently the executable is built first, and then the tests are build. We should switch the order, and have the build process as a whole fail if a unittest fail (no executable built).
2. The error message for a failed test is barely visible, and not enough informative. It doesn't tell you the test name and the actual assertion that failed, making it impossible to debug.

These are all features we had before.

CMakeLists.txt
```diff
@@ -0,0 +1,223 @@
...
+add_executable(${MARABOU_REGRESS} ${REGRESS_DIR}/main_regress.cpp)
+
+
+# Build regress 
+target_link_libraries(${MARABOU_REGRESS} PRIVATE ${MARABOU_LIB})
+
+target_include_directories(${MARABOU_REGRESS} PRIVATE ${LIBS_INCLUDES})
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/logs)
+file(COPY ${ACAS_REGRESS_NET} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}) 
+file(COPY "${REGRESS_DIR}/lp_feasible_1.mps" DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
guykatzz: Is it possible to move this logic (listing of specific files) to the CMakeLists under regress?

guykatzz:
Remove regress/Makefile?

guykatzz:
Also, please set the unit tests to run by default - we can change this in the future if we see that they become too many.

PR #177, opened 2019-09-18
add macro to only call sched_getcpu on linux

anwu1219:


guykatzz:
I'm not sure "common/CrossPlatform.h" is the place for this code - maybe just for the definition of a LINUX( ... ) macro. The rest of the code is very specific to DnC, so should be included within the DnC files. Or, alternatively, you can rename it to something more general (getCPUData, or something), and leave it under common.

src/engine/DnCManager.cpp
```diff
@@ -37,7 +37,10 @@ void DnCManager::dncSolve( WorkerQueue *workload, std::shared_ptr<Engine> engine
                            unsigned threadId, unsigned onlineDivides,
                            float timeoutFactor, DivideStrategy divideStrategy )
 {
-    log( Stringf( "Thread #%u on CPU %u", threadId, sched_getcpu() ) );
+    unsigned cpuId = 0;
+    getCPUId( cpuId );
+    log( Stringf( "Thread #%u on CPU %u", threadId, cpuId ) );
```
guykatzz: Add \n after %u.

Apart from that, LGTM.

guykatzz:
LGTM

PR #178, opened 2019-09-20
remove assumption that merged and fixed variables are disjoint 

yuvaljacoby:
It's a minor change.
In Preprocessor.cpp we assume that fixedVariables and mergedVariables are disjoint (preprocess function).

I tried an example that it wasn't the case and then number of variables overflowed and the query got stuck. 
Not the most elegant solution but fixed my problem...


guykatzz:
I don't really understand this fix. You're suggesting that the "variableCount" can be negative, and that it should be set to 0 instead? We should add a more general fix, as it seems this phenomenon can happen also when variableCount is positive.

yuvaljacoby:
You are probably right.
This fix is a patch and not what we actually need to do.
I pushed code that reproduce the problem, I prefer not to fix it right now (after finishing CMake I can do it). In the mean time should I also open an issue?

src/engine/Makefile
```diff
@@ -44,13 +44,8 @@ vpath %.cpp $(ENGINE_REAL_DIR)
...
 
 SOURCES += \
-	DnCMarabou.cpp \
-	Marabou.cpp \
-	OptionParser.cpp \
-	Options.cpp \
-	main.cpp \
-
-TARGET = marabou.elf
+		   variables_test.cpp
```
guykatzz: I think this cpp file is missing

guykatzz:
For the failing example, did you run in DEBUG mode? Did the assertion at the end of Preprocessor::preprocess go off?

yuvaljacoby:
Didn't run in debug mode, didn't know about the DEBUG_ON flag.
After adding it the assert raises exception.

guykatzz:
Please try now

yuvaljacoby:
Works, Thanks.
Cleaned the pull request, I think we can squash and merge.

guykatzz:
Great.
Before we merge, please add your previously failing example as either a
unit test for Preprocessor.h, or as a regression test, and see that it
still fails if you comment out the new function call.

yuvaljacoby:
done 
did also a small fix in sec/common/debug , we need the three dots there in order to use variables.

The default compilation of the tests does not have the DEBUG_ON flag, should we change it?

guykatzz:
Yes, lets have DBEUG_ON for the tests.

PR #185, opened 2019-10-01
Consistent output format between DnC mode and the sequential mode

anwu1219:
Addressing https://github.com/guykatzz/Marabou/issues/154. Make the results printed out by DnCMarabou and Marabou be in the same format.

### DIFF
--- a/src/engine/DnCManager.cpp
+++ b/src/engine/DnCManager.cpp
@@ -230,6 +230,7 @@ String DnCManager::getResultString()
 
 void DnCManager::printResult()
 {
+    std::cout << std::endl;
     switch ( _exitCode )
     {
     case DnCManager::SAT:
--- a/src/engine/DnCManager.cpp
+++ b/src/engine/DnCManager.cpp
@@ -234,7 +234,7 @@ void DnCManager::printResult()
     {
     case DnCManager::SAT:
     {
-        std::cout << "SAT" << std::endl;
+        std::cout << "SAT\n" << std::endl;
 
         ASSERT( _engineWithSATAssignment != nullptr );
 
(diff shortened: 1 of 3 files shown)


src/engine/DnCManager.cpp
```diff
@@ -264,19 +263,19 @@ void DnCManager::printResult()
...
     }
     case DnCManager::UNSAT:
-        std::cout << "DnCManager::solve UNSAT query" << std::endl;
+        std::cout << "UNSAT" << std::endl;
         break;
     case DnCManager::ERROR:
-        std::cout << "DnCManager::solve ERROR" << std::endl;
+        std::cout << "ERROR" << std::endl;
         break;
     case DnCManager::NOT_DONE:
```
ahmed-irfan: is this needed now? since it is called after the DnCManager finishes....

src/engine/DnCManager.cpp
```diff
@@ -264,19 +263,19 @@ void DnCManager::printResult()
...
-        std::cout << "DnCManager::solve NOT_DONE" << std::endl;
+        std::cout << "NOT_DONE" << std::endl;
         break;
     case DnCManager::QUIT_REQUESTED:
```
ahmed-irfan: ditto

ahmed-irfan:
a couple of minor points...
other than that LGTM

src/engine/DnCManager.cpp
anwu1219: In the sequential Marabou, we also print out result string other than SAT, UNSAT, and UNKNOWN, such as ERROR and TIMEOUT.
https://github.com/guykatzz/Marabou/blob/c05a9a870b9deb5d5efbe0b2ae564f4598f5a4a8/src/engine/Marabou.cpp#L122

We might need to discuss what should Marabou print out as results as a whole.

PR #187, opened 2019-10-08
Changes for Windows that also affect Linux

kjulian3:
This is the first of two pull requests so that Marabou can be compiled on a Windows machine using Visual Studio. This PR contains changes to files that affect compilation on all machines, whereas the other PR will only contain changes that affect compilation on a Windows machine.

There are four types of changes:
1. Adding a comma before "..." in macro argument list
2. Removing instances where arrays are initialized with variable length
3. Changing the divide-by-zero test to use a variable instead of hard-coded zero
4. "xor" to "^"

These are all small changes needed to compile on Windows with Visual Studio but are not Windows specific. I'm able to run tests and maraboupy successfully on my Linux machine with these changes added.

### DIFF
--- a/src/input_parsers/AcasNeuralNetwork.cpp
+++ b/src/input_parsers/AcasNeuralNetwork.cpp
@@ -16,6 +16,7 @@
 #include "AcasNeuralNetwork.h"
 #include "AcasNnet.h"
 #include "FloatUtils.h"
+#include "Vector.h"
 
 #include <iostream>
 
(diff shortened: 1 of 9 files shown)


guykatzz:
I pushed a few minor changes. Please see that it still works on Windows - if so, I think you can go ahead and merge.

src/input_parsers/AcasNeuralNetwork.cpp
```diff
@@ -76,8 +76,8 @@ void AcasNeuralNetwork::evaluate( const Vector<double> &inputs, Vector<double> &
     double *input( inputVector.data() );
     double *output( outputVector.data() );
 
-    memset( input, 0, num_inputs( _network ) );
-    memset( output, 0, num_outputs( _network ) );
+    std::fill_n( input, num_inputs( _network ), 0 );
+    std::fill_n( output, num_outputs( _network ), 0 );
```
kjulian3: This actually creates a seg fault (for both Linux and Windows).

The argument outputs is an empty vector, which we append to on line 96. So outputs.size() is 0, which means the double array output is not the right size. So when fill_n tries to fill the array, it fills more than is allocated.

The simplest fix might be to use num_outputs( _network ) instead of outputs.size(). That is working for me now

guykatzz:
Can you please check again now? I prefer to use the outputSize parameter,
which is used again later in that method.

kjulian3:
The "Allocations" commit works fine. We can also simplify this by adding a constructor in Vector.h that takes in a second argument - the initial value for all members of the vector. 

I also added simple tests to Test_Vector.h for the new constructors we've added. The code compiles and passes all tests on Linux and Windows (I will submit the changes that affect only Windows compilation in the next PR)

kjulian3:
If this looks good to you, feel free to merge

PR #189, opened 2019-10-14
Building on Windows

kjulian3:
This pull request contains Windows-specific changes needed to build Marabou on a Windows machine with Visual Studio.

There are a few types of changes made to the code:
1. New files to replace standard c++ headers that are not available with Visual Studio. Some of these files come from existing sources and have a header block at the top of the file, so I wasn't sure if we should use the Marabou header block or not and just left them as is.
2. Source file changes that are surrounded by "#ifdef _WIN32" statements so they only affect compilation on windows machines.
3. Edits to CMakeLists. Most edits use "if (MSVC)" statements so they only affect windows machines. A few edits were also made to put file paths inside quotes.
4. README edits for instructions on building and using Marabou on a Windows machine.
5. Minor typo fixes in TimeUtils.cpp (secodnsSinceEpoch -> secondsSinceEpoch ) and Engine.h (cauciously -> cautiously). I noticed the typos while working on this PR and added them in since they are so minor.

I've tested this branch successfully on both Windows and Linux machines. It would be great to test this on other Windows machines before merging to make sure I am not missing any important steps for building on Windows.


CMakeLists.txt
```diff
@@ -68,42 +68,53 @@ if(${BUILD_PYTHON})
...
+    set(Boost_NAMESPACE libboost)
+endif()
 find_package(Boost COMPONENTS program_options)
 # Find boost
 if (NOT ${Boost_FOUND})
     # bash file that downloads and install boost 1_68_0, the name need to match
     # BOOST_DIR variable
-    execute_process(COMMAND ${TOOLS_DIR}/download_boost.sh)
+    if(MSVC)
+        execute_process(COMMAND ${TOOLS_DIR}/download_boost.bat)
```
yuvaljacoby: This happens a couple of times, maybe will be easier to set an SCRIPT_EXTENSION variable and set it to sh / bat according to the platform

CMakeLists.txt
```diff
@@ -68,42 +68,53 @@ if(${BUILD_PYTHON})
...
+if (MSVC)
+    set(Boost_NAMESPACE libboost)
+endif()
 find_package(Boost COMPONENTS program_options)
 # Find boost
 if (NOT ${Boost_FOUND})
     # bash file that downloads and install boost 1_68_0, the name need to match
     # BOOST_DIR variable
-    execute_process(COMMAND ${TOOLS_DIR}/download_boost.sh)
+    if(MSVC)
```
yuvaljacoby: Actually I think we should check if it's WIN32 (aka a windows terminal) or not.
MSVC tells you whether we are using cl (visual studio compiler) or not

tools/download_boost.bat
```diff
@@ -0,0 +1,18 @@
+@echo off
+SET mydir=%~dp0
+SET curdir=%cd%
+
+CD %mydir%
+
+REM TODO: add progress bar, -q is quiet, if removing it the progress bar is in multiple lines
+ECHO Downloading Boost
+wget -q  https://dl.bintray.com/boostorg/release/1.68.0/source/boost_1_68_0.tar.gz
```
yuvaljacoby: I guess we should use curl and not wget for windows, wget is not supported out-of-box...

README.md
```diff
@@ -76,46 +81,82 @@ For example to run all unit tests execute in the build directory:
...
+cd build
+cmake .. -G"Visual Studio 15 2017 Win64" -DBUILD_PYTHON=ON
+cmake --build . --config Release
+```
 This process will produce the binary file and the shared library for the Python 
-API, the shared library will be at maraboupy folder
-
+API. The shared library will be in the maraboupy folder for Linux and MacOS. 
+On Windows, the shared library is written to a Release subfolder in maraboupy, 
+so you will need to move the Release/\*pyd file to the maraboupy folder:
```
yuvaljacoby: We should keep the same behavior, you can easily copy the target with CMake command (talk to me if needed).

But I actually was rethinking about it and not sure we should copy it at all, @guykatzz  @mykelk  what do you think? when building marabou we create the Python API shared library, should we copy it to the maraboupy (and make it easy to use), or keep it in the build folder and keep the build "clean" (i.e. the build does not create any thing outside the dir)

README.md
```diff
@@ -76,46 +81,82 @@ For example to run all unit tests execute in the build directory:
 ```
 ctest -L unit
 ```
+### Build Instructions for Windows using Visual Studio
+
+First, install Visual Studio and select the "Desktop development with C++" workload.
```
yuvaljacoby: Maybe specify to install visual studio 2017 or later?

README.md
mykelk: Yeah, keeping it clean might be best.

CMakeLists.txt
```diff
@@ -68,42 +68,53 @@ if(${BUILD_PYTHON})
     endif()
 endif()
 
-if(MSVC)
-    # TODO: TEST this
-    set(COMPILE_FLAGS  /W4 /WX) 
-else()
```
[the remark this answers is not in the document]
kjulian3: I removed it because I was seeing warnings that were not present when compiling on Linux. These warnings include "possible loss of data during conversion from 'size_t' to 'unsigned int'", "multiple copy constructors specified", "declaration of 'i' hides previous local declaration", "'*/' found outside of comment", etc. 

We may want to look at some of these warnings more closely, but others like "'*/' found outside of comment" are because files like SparseLUFactorization.cpp contain inline "/*  */" style comments. We can add compiler flags to tell Visual Studio to ignore certain kinds of warnings that we don't care about, so it would be possible to add /WX while ignoring unnecessary warnings.

I was thinking of submitting another pull request after this that addresses warnings generated by Visual Studio. Some of the warnings may require code changes that affect building on Linux too, so I didn't want to make this PR too complicated for now and took out the /W4 and /WX flags.

CMakeLists.txt
```diff
@@ -68,42 +68,53 @@ if(${BUILD_PYTHON})
...
-    set(MEMORY_FLAGS -fsanitize=address -fno-omit-frame-pointer -O1)
+    if(NOT MSVC)
+        set(MEMORY_FLAGS -fsanitize=address -fno-omit-frame-pointer -O1)
+    endif()
 endif()
 
-set(DEBUG_FLAGS ${COMPILE_FLAGS} ${MEMORY_FLAGS} -g -DDEBUG_ON) 
-set(CXXTEST_FLAGS ${DEBUG_FLAGS}  -Wno-ignored-qualifiers)
+if (NOT MSVC)
+    set(DEBUG_FLAGS ${COMPILE_FLAGS} ${MEMORY_FLAGS} -g -DDEBUG_ON)
```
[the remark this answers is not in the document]
kjulian3: Oh I see, I'll fix that then.

CMakeLists.txt
kjulian3: Yes that makes sense and would make this cleaner. I'll fix this

tools/download_boost.bat
kjulian3: Yes that would make sense to remove an extra step for users. I'll try it out with curl to make sure that would work ok and push a commit if so

CMakeLists.txt
kjulian3: I'm not sure if it will make too much difference. The download_boost.bat uses the msvc toolset to build the program_options library, so it's currently assumed that MSVC is going to be used from a windows machine.

tools/download_boost.bat
kjulian3: Fixed with 5646b0a

CMakeLists.txt
kjulian3: Fixed with ea38fbd

CMakeLists.txt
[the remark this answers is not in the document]
kjulian3: Fixed with ea38fbd

README.md
guykatzz: I agree with Mykel. We can keep it in the build folder until there's a reason to do otherwise.

src/engine/Simulator.cpp
```diff
@@ -19,6 +19,10 @@
 #include "MarabouError.h"
 #include "Simulator.h"
 
+#ifdef _WIN32
+#include "time.h"
```
guykatzz: Is time.h a system file? If so, can we change to <time.h>?

kjulian3: Yes this is a system file. Fixed with 7c28456.

README.md
kjulian3: Updated the readme with 5cebe1c

README.md
kjulian3: How would we be able to run maraboupy from python and not move the python dynamic link library file to the maraboupy folder?

yuvaljacoby: The correct way to do it is:
install the python library for users (make install, that will put the .so in /usr/lib )
edit the python path for developers (add build/lib/*.so to the PYTHONPATH variable), this is not a permanent solution because need to edit the PYTHONPATH every time opening a new terminal.

I think that for this pull request we can keep the situation as is, and I will open a new issue and will solve it in the near future, sounds good?

kjulian3: I see, and currently users can edit .bashrc (or environmental variables on Windows) so that PYTHONPATH is already edited for every new terminal, which is a quick one time change.

I agree that a separate issue and pull request for this makes sense.

maraboupy/MarabouCore.cpp
```diff
@@ -37,6 +36,12 @@
 #include "ReluConstraint.h"
 #include "Set.h"
 
+#ifdef _WIN32
+#include <T/winunistd.h>
```
guykatzz: Files that are part of the project should be included with " ", and system files with < >

regress/RegressUtils.h
```diff
@@ -27,7 +27,11 @@
 
 #include "MStringf.h"
 #include "TimeUtils.h"
+#ifdef _WIN32
+#include <T/winunistd.h>
```
guykatzz: Files that are part of the project should be included with " ", and system files with < >

src/common/SignalHandler.cpp
```diff
@@ -36,12 +36,21 @@ void SignalHandler::registerClient( Signalable *client )
 
 void SignalHandler::initialize()
 {
+
+#ifdef _WIN32
+    signal( SIGINT, got_signal );
+    signal( SIGTERM, got_signal );
+    signal( SIGABRT, got_signal );
+
```
guykatzz: Remove empty line please, for symmetry

src/common/TimeUtils.cpp
```diff
@@ -19,7 +19,24 @@
 struct timespec TimeUtils::sampleMicro()
 {
     struct timespec now;
+#ifdef _WIN32
+	LARGE_INTEGER count;
+
+	if (g_first_time)
```
guykatzz: Style: space before/after ( and )

src/common/TimeUtils.h
```diff
@@ -19,7 +19,14 @@
 #include "MString.h"
 
 #include <time.h>
+
+#ifdef _WIN32
+#include <winsock2.h>
+static BOOL g_first_time = 1;
```
guykatzz: @AleksandarZeljic Is this the style we decided on for global variables?

AleksandarZeljic: @guykatzz There is nothing in the coding guidelines about style of global variables, so I don't think we made a decision. Are you for or against this style? I know we decided to drop the prefix in some cases, where that made sense.

AleksandarZeljic: SGTM, I can add it to the Coding Guidelines.

kjulian3: Pushed a commit to use camel case for the global variables

guykatzz:
LGTM

PR #191, opened 2019-10-16
Optimize floatutils

yuvaljacoby:
To validate the improvement I run the following networks:
acas_2_5 property 2 (10 minutes run, SAT), improvement of ~6.8% over 5 experiments 
acas_3_3 property 3 (8 minutes run, UNSAT), improvement of ~6.6% over 5 experiments 
acas_2_8 property 4 (3 minutes run, UNSAT), improvement of ~7.4% over 5 experiments 

Most of the most of the speedup comes from the inline (the size of the executable is similar).
The reduce of operations gives another ~0.5 improvement, in isPositive / isNegiative the compiler can't do this optimization because the epsilon is not static, so that actually removes two operations.
In isZero, we replace a compression into multiply and 2 additions  which is supposed to be faster


src/common/FloatUtils.h
```diff
@@ -18,21 +18,37 @@
 
 #include "GlobalConfiguration.h"
 #include "MString.h"
+#include "Debug.h"
```
guykatzz: Cosmetics: sort include lines alphabetically

src/common/FloatUtils.h
```diff
@@ -18,21 +18,37 @@
...
 class FloatUtils
 {
 public:
+    inline static bool isZero( double x, double epsilon = GlobalConfiguration::DEFAULT_EPSILON_FOR_COMPARISONS )
+    {
+        ASSERT( epsilon > 0 ); 
+        double lower = -epsilon;
+        double upper = epsilon;
+        return ( x - upper ) * ( x - lower ) <= 0;
+    }
```
guykatzz: cosmetics: add empty line between functions

guykatzz:
Changes look okay. IsNegative and IsPositive definitely seem better; regarding isZero, I don't know, so an extensive comparison is needed. Can you start a cluster experiment on the entire ACAS Xu benchmark suite? At least properties 1-4.

yuvaljacoby:
I did the test separate, checked only the inline and then inline + the code changes (all code changes together).
most of the performance come from the inline, and a bit more from the code changes.

BTW, in gcc5 (and 7) it's enough to move the implementations to the .h file and the compiler is making the function inline automatically (so we can remove the inline keyword, a matter of style) 
didn't check clang

ahmed-irfan:
@yuvaljacoby why not put the full implementation of floatutils in the header file then....

guykatzz:
Hi Yuval,
What's the status of this pull request?

yuvaljacoby:
I am sorry, forgot about it.
Will get back to it in the next couple of days

ahmed-irfan:
@yuvaljacoby here is the comparison you asked for:

[p4.txt](https://github.com/NeuralNetworkVerification/Marabou/files/4434945/p4.txt)
[p3.txt](https://github.com/NeuralNetworkVerification/Marabou/files/4434947/p3.txt)
[p2.txt](https://github.com/NeuralNetworkVerification/Marabou/files/4434948/p2.txt)
[p1.txt](https://github.com/NeuralNetworkVerification/Marabou/files/4434949/p1.txt)

yuvaljacoby:
Summering for future reference (only avg cpu):
p1:
original: 140421.1
inline:    139222.3
p2:
original: 63249.9
inline:    61641.4
p3:
original: 23690.4
inline:    22135.7
p4:
original: 10303.6
inline:    9252.6

guykatzz:
LGTM

PR #194, opened 2019-10-23
Unit test for DnCWorker

anwu1219:
1. Refactor the DnCWorker to facilitate unit tests
The output of this refactored version is consistent with the current version (tested on AcasXU properties 1-4).
2. Add unit test for DnCWorker


src/engine/Engine.cpp
```diff
@@ -1811,6 +1811,15 @@ bool Engine::shouldExitDueToTimeout( unsigned timeout ) const
     return _statistics.getTotalTime() / MILLISECONDS_TO_SECONDS > timeout;
 }
 
+void Engine::reset()
+{
+    Statistics *statistics = new Statistics();
```
guykatzz: This looks like a memory leak. Maybe just do "Statistics statistics;", and move this to be inside resetStatistics().

src/engine/Engine.cpp
```diff
@@ -1811,6 +1811,15 @@ bool Engine::shouldExitDueToTimeout( unsigned timeout ) const
...
+    resetStatistics( *statistics );
+    clearViolatedPLConstraints();
+    resetSmtCore();
+    resetBoundTighteners();
+    resetExitCode();
+}
 
 void Engine::resetStatistics( const Statistics &statistics )
 {
```
guykatzz: I noticed that we're setting the active entry strategy in this method. This doesn't seem to belong here.

src/engine/DnCManager.cpp
```diff
@@ -147,14 +150,15 @@ void DnCManager::solve( unsigned timeoutInSeconds )
 
     // Wait until either all subQueries are solved or a satisfying assignment is
     // found by some worker
-    while ( _numUnsolvedSubQueries.load() > 0 &&
-            !shouldQuitSolving.load() &&
-            !_timeoutReached )
+    while ( !shouldQuitSolving.load() && !_timeoutReached )
```
guykatzz: This seems a little redundant. Why check _timeoutReached, if when it is reached we set shouldQuitSolving to true? Seems like it would suffice to check shouldQuitSolving

src/engine/DnCWorker.cpp
```diff
@@ -60,132 +60,122 @@ void DnCWorker::setQueryDivider( DivideStrategy divideStrategy )
...
-                for ( auto &newSubQuery : subQueries )
+                if ( !_workload->push( std::move( newSubQuery ) ) )
                 {
-                    if ( !_workload->push( std::move( newSubQuery ) ) )
-                    {
-                        ASSERT( false );
-                    }
-
-                    *_numUnsolvedSubQueries += 1;
+                    ASSERT( false );
```
guykatzz: Should we perhaps throw an error here? In non-DEBUG mode, the execution won't stop here.

src/engine/DnCWorker.cpp
```diff
@@ -60,132 +60,122 @@ void DnCWorker::setQueryDivider( DivideStrategy divideStrategy )
...
             }
-            else if ( result == Engine::SAT )
+        else
+        {
+            // We must quit solving if the result is not UNSAT or TIMEOUT
+            *_shouldQuitSolving = true;
+            if ( result == IEngine::SAT )
             {
                 // If SAT, set the shouldQuitSolving flag to true, so that the
                 // DnCManager will kill all the DnCWorkers
```
guykatzz: Comment says we should set the quit flag, but do we do that?

guykatzz:
Looks good, overall. I polished a bit and left a few comments.

src/engine/Engine.cpp
anwu1219: Ok, I removed that line

anwu1219:
I'll run the modified version on the AcasXU properties again to make sure I didn't break anything accidentally.

src/engine/DnCWorker.cpp
anwu1219: Throwing error here.

src/engine/DnCWorker.cpp
anwu1219: We set the flag before case SAT/ERROR/NOT_DONE.
The comment in the branches might be misleading so I've updated them.

Issue #193, opened 2019-10-22
Problem with MarabouNetwork.loadQuery()


darioguidotti:
Hello,

I have tried to use MarabouNetwork.loadQuery() in order to load my
verification query but it gives me the following error:

> ~/Marabou-master/maraboupy/MarabouNetwork.py in loadQuery(self, filename, verbose, timeout)
>     220         """
>     221         #ipq = self.getMarabouQuery()
> --> 222         ipq = MarabouCore.loadQuery(filename)
>     223         vals, stats = MarabouCore.solve(ipq, filename, timeout=0)
>     224         if verbose:
> 
> AttributeError: module 'maraboupy.MarabouCore' has no attribute 'loadQuery'

Do you have any suggestions?
Thank you!

kjulian3:
@clazarus, I think you added loadQuery to maraboupy in #61. Have you tested that method? It looks like loadQuery was never added to MarabouCore.cpp like saveQuery was, so I think a fix is needed. Do you have time to look at this? I can try to fix this if not.

darioguidotti:
Hello,

Thank you for your help!

Right now loadQuery appears to return a "MarabouCore.InputQuery": how can I use it alongside a 
MarabouNetworkNNet to check the property of the network?

Thank you again!

kjulian3:
The old loadQuery would try to load the query and solve it, so the function name was misleading. Now, loadQuery returns a MarabouCore.InputQuery object, which can be solved by calling Marabou.solve_query. So now if you have a serialized inputQuery file, you can load and solve it with
```
from maraboupy import Marabou
ipq = Marabou.load_query(filename)
vals, stats = Marabou.solve_query(ipq)
```
Are you checking an nnet file, or a serialized inputQuery file?

darioguidotti:
Hi,

Actually I need to check the union of both, i.e., I have a network loaded using .nnet file and then I have a .txt file with the properties of interest encoded following [the Marabou specifications](https://github.com/guykatzz/Marabou/wiki/Marabou-Input-Formats). I am not sure how I can check the network against the property as in the command line example:

> ./build/Marabou resources/nnet/acasxu/ACASXU_experimental_v2a_2_7.nnet resources/properties/acas_property_3.txt

Thank you for your time!

kjulian3:
[...]

PR #195, opened 2019-10-24
Changes to fix save/loadQuery functions

kjulian3:
This PR addresses #193 and enables loadQuery to work properly on c++ and python. Changes made include:
1. Reorganizing the python API so queries can be loaded and solved without needing a MarabouNetwork object.
2. Added unit test for loadQuery function, which ensures that saving and loading an InputQuery will result in the same InputQuery.
3. Adjusting the saveQuery method so that marked input/output variables are saved, and so that lower/upper bounds are written separately. Users may only specify an upper or lower bound for a variable, so we can't write the bounds together on the same line.
4. The trim() function in String has a bug that occurs when the input string has only one character. If that's the case, lastNonSpace will be "updated" to 0 during the one pass through the loop, and then an empty string gets returned instead of the single character. A minor fix was made that allows single character non-space strings to be returned correctly. This was caught during the loadQuery unit test.
5. Minor adjustments to the python API, such as adding verbosity as an input so Marabou can be easily run silently in python.


kjulian3:
@yuvaljacoby I'm adding a unit test so I had to edit CMakeLists and add a test header file. I followed an existing unit test as a guide. Did I set it up correctly, or is there anything to fix?

@guykatzz Besides save/load query functions, I made a minor change to MString trim function that could affect other times trim is used. The changes I made satisfy the unit tests, but I want to make sure I'm not unintentionally messing up something else. Also, should I add the header/footer comment blocks to the new test header file, or is that something done automatically?

src/engine/InputQuery.cpp
```diff
@@ -290,26 +290,51 @@ void InputQuery::saveQuery( const String &fileName )
...
+    }
+
+    // Output Variables
+    queryFile->write( Stringf( "\n%u", getNumOutputVariables() ) );
+    i = 0;
+    for ( const auto &outVar : getOutputVariables() )
+    {
+        queryFile->write( Stringf( "\n%u,%u", i, outVar ) );
+        ++i;
+    }
```
yuvaljacoby: * To keep consistency of the file, add description (Number of input variables etc...)
* I would also assert that _i_ is the same as the getNumOutputVariables (otherwise the load method will be inconsistent) same thing for input

src/query_loader/QueryLoader.cpp
```diff
@@ -17,47 +17,76 @@
...
+        throw MarabouError( MarabouError::FILE_DOES_NOT_EXIST, Stringf( "File %s not found.\n", fileName.ascii() ).ascii() );
     }
 
     InputQuery inputQuery;
     File input = File( fileName );
     input.open( File::MODE_READ );
 
     unsigned numVars = atoi( input.readLine().trim().ascii() );
-    unsigned numBounds = atoi( input.readLine().trim().ascii() );
+    unsigned numLowerBounds = atoi( input.readLine().trim().ascii() );
```
yuvaljacoby: make sure atoi is positive? 
I am not sure it's important, we are not writing defensive code and stuff might overflow, just popped to me (an easy change can be using a define that throes exception if it's not positive) ...

src/query_loader/tests/Test_QueryLoader.h
```diff
@@ -0,0 +1,125 @@
...
+        Equation equation2;
+        equation2.addAddend( -1.0, 5 ); // Equation output
+        equation2.addAddend( -1.0, 3 ); // Weighted equation input
+        equation2.addAddend( 1.0, 4 );  // Weighted equation input
+        equation2.setScalar( 0.5 );     // Equation bias
+        inputQuery.addEquation( equation2 );
+
+        // Save the query and then reload the query
+        inputQuery.saveQuery( "TEST_QUERY.txt" );
+        InputQuery inputQuery2 = QueryLoader::loadQuery( "TEST_QUERY.txt" );
```
yuvaljacoby: delete the file if exists in cleanup (and maybe move the name to define so it will be easy to change)

src/engine/InputQuery.cpp
kjulian3: Added the assertion and better descriptions

src/query_loader/QueryLoader.cpp
kjulian3: I can add the define that would throw an exception, but it makes the code messier. There are also instances in the input parsers where the output of atoi is cast to unsigned without checking if the integer is negative. Do we want to add these protective checks when using atoi?

src/query_loader/tests/Test_QueryLoader.h
kjulian3: Added a line to remove temp file during clean up

src/query_loader/QueryLoader.cpp
yuvaljacoby: Don't see why it's messy (will be just to call atoi_safe instead of atoi), but I agree, @guykatzz  what do you think?

src/engine/InputQuery.cpp
yuvaljacoby: Oh no I understand that my comment about the description was kind of stupid, sorry.
I thought that we write to the file: NumberOfVariables: XXX 
my mistake, thanks for adding the asserts.

src/query_loader/QueryLoader.cpp
kjulian3: If you do something like `#define atoi_safe(a) a>=0?atoi(a):throw "Converting negative integer to unsigned"`, with the input being `input.readLine()`, the macro will do a double evaluation and call `input.readLine()` twice, since `a` appears twice in the macro. One way around this would be to save the output of `input.readLine()` to a variable before using the macro. Or maybe a better way would be to add functions to MString to do atoi and safe_atoi so the function could be defined once and called on any MString variable. This could be a separate issue / pull request?

yuvaljacoby:
Well you can also do:
atoi_safe(a) b = atoi(a); b>=0?b:throw

Since you said it's already a problem that we have, I approved the pull
request.
I guess that one day we should start writing defensive code, doesn't have
to start today...

src/query_loader/QueryLoader.cpp
guykatzz: I don't feel very strongly either way. I think atoi() in general has been deprecated and should not be used, but as Kyle mentioned this isn't enforced in our code. I think in this case it would be enough to change the types to int and do an ASSERT that they are positive, but any solution you pick is fine.

guykatzz:
Hi Kyle,
Changes look good to me, and thanks for tracking and taking care of the MString bug! 
My only comment is about the unit test: we usually try to make tests not create files, open sockets, etc. The way we do this is by using CxxTest Mock infrastructure. I will change this and push the changes.

src/query_loader/QueryLoader.cpp
kjulian3: If we use ASSERT, does that only effect the code if DEBUG_ON is set? So if a user is running in release mode, the ASSERT macro has no effect and allows the program to continue even if negative ints are cast to unsigned ints?

guykatzz: That's correct. If we wanted this to happen without DEBUG_ON, we would need to do an "if ... throw" instead of an assert.

kjulian3: Ok makes sense. Do you have preference either way? Should we leave it as is and open an issue to address use of atoi to read unsigned ints throughout Marabou and continue discussion on the issue?

guykatzz: Yes, that makes sense.

kjulian3:
Everything looks good to me, and thanks for the explanation for how to use Mock for unit tests. Am I good to merge?

guykatzz:
Go for it...

PR #199, opened 2019-11-02
Regression framework

yuvaljacoby:
The commit is not big is it seems to be.
Deleted a bunch of files from the regress folder

I divided the regression tests into up to 5 levels.
Currently filled only levels 0-2 where level 0 is only for sanity test (only one test per scenario).

in travis I am running level 0 on all platforms, and levels 1-2 only on gcc.
please make sure reviewing the README, I am not sure that I wrote it clear enough

Moved some cleaning from Enginge to InputQuery, solved memory leak in one of the system tests


regress/regress1/CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
+
+marabou_add_coav_test(1 "reluBenchmark1.30941200256s_UNSAT.nnet" UNSAT) 
+marabou_add_coav_test(1 "reluBenchmark0.94518494606s_SAT.nnet" SAT)
```
guykatzz: How did we pick these specific ones? Because they terminate quickly?

src/basis_factorization/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ target_include_directories(${MARABOU_TEST_LIB} PUBLIC "${CMAKE_CURRENT_SOURCE_DI
  
 set (BASIS_FACTORIZATION_TESTS_DIR "${CMAKE_CURRENT_SOURCE_DIR}/tests")
 macro(basis_factorization_add_unit_test name)
-    marabou_add_test(${BASIS_FACTORIZATION_TESTS_DIR}/Test_${name} basis_factorization TRUE "unit")
+    marabou_add_test(${BASIS_FACTORIZATION_TESTS_DIR}/Test_${name} basis_factorization TRUE TRUE "unit")
```
guykatzz: This "TRUE TRUE" business isn't clear. Does CMake have something like an enum? So we could pass, e.g., USE_MOCK_ENGINE instead of TRUE?

src/input_parsers/AcasNnet.cpp
```diff
@@ -28,7 +30,7 @@ AcasNnet *load_network(const char* filename)
 
     if (fstream == NULL)
     {
-        return NULL;
+        throw InputParserError( InputParserError::FILE_DOESNT_EXIST);
```
guykatzz: add space before )

src/system_tests/Test_acas.h
```diff
@@ -0,0 +1,171 @@
...
+        for ( unsigned i = 0; i < 5; ++i )
+        {
+            unsigned variable = acasParser.getOutputVariable( i );
+            double newError = FloatUtils::abs( outputs[i] - inputQuery.getSolutionValue( variable ) );
+            if ( FloatUtils::gt( newError, maxError ) )
+              maxError = newError;
+        }
+
+        TS_ASSERT( FloatUtils::lt( maxError, 0.00001 ) );
+
```
guykatzz: remove empty lines in this file

src/system_tests/Test_lp.h
```diff
@@ -0,0 +1,159 @@
...
+ ** [[ Add lengthier description here ]]
+
+**/
+
+#include <cxxtest/TestSuite.h>
+
+#include "Engine.h"
+#include "InputQuery.h"
+
+
```
guykatzz: empty lines in this file, too

src/system_tests/Test_max.h
```diff
@@ -0,0 +1,213 @@
...
+
+        Equation equation3;
+        equation3.addAddend( 1, 2 );
+        equation3.addAddend( 1, 4 );
+        equation3.addAddend( -1, 5 );
+        equation3.addAddend( 1, 8 );
+        equation3.setScalar( 0 );
+        inputQuery.addEquation( equation3 );
+
+		MaxConstraint *max1 = new MaxConstraint( 5, Set<unsigned>( { 0, 2, 3 } ) );
```
guykatzz: indentation

src/system_tests/Test_mps.h
```diff
@@ -11,55 +11,62 @@
...
-
         Engine engine;
         if ( !engine.processInputQuery( inputQuery ) )
         {
-            struct timespec end = TimeUtils::sampleMicro();
-            restoreOutputStream( outputStream );
-            printFailed( "mps_lp_feasible_1", start, end );
-            return;
+            // got infeasible in preprocess stage
+            TS_ASSERT( 1 );
```
guykatzz: TS_ASSERT( 1 )? Maybe 0?

yuvaljacoby: It's infeasible test, I don't want to fail the test here, just wanted clarify that it's a good situation.
seems from the comment that I didn't do it well enough, any other idea how to say that this is a good situation (that the preprocessing failed)

src/basis_factorization/CMakeLists.txt
yuvaljacoby: Can't see how to do it nice with enum (CMake has enums)
Changed the literal TRUE to a variable which explains the purpose, is this better?

yuvaljacoby: Oh I got you, you mean to pass to the function something like: USE_MOCK_ENGINE::TRUE
That is not possible, the closest thing to enum in CMake is to check if a variable is within a set of values.
Actually CMake type is always a string, some commands can treat a variable as path / bool but that's it: https://cgold.readthedocs.io/en/latest/tutorials/variables/regular.html#types-of-variable

guykatzz: Ok

regress/regress1/CMakeLists.txt
yuvaljacoby: I remember answering this but don't find the answer now...
Yes I am not familiar with the different networks, so I choose everything by the termination time.
In regress0, the fastest example I found per case (SAT and UNSAT for every network type)
In regess1, Some acasxu networks, most of coav (since they are pretty fast) 
and In regress2 some more of acasxu networks.

Can you think on something clever to do?

guykatzz:
This sounds like a good strategy.

PR #203, opened 2019-11-19
Choose python version

yuvaljacoby:
Add option to choose between python 2 and 3 (where default is 3)
If only one python version is installed will choose that, if both are installed will use the default.

Other then that fixed the tests.py example (after introducing MarabouOptions) which broke it....


GuyKatzHuji:
Looks good to me.

PR #213, opened 2020-01-20
Onnx maraboupy parser

kjulian3:
This pull request introduces an ONNX parser for Marabou. The format is very similar to parsing tensorflow models, though a few things are different. ONNX does not have functions for easily computing the output shape of intermediate layers, so the shapes of layers in the graph must be inferred from input shapes, constant shapes, and operations.

I tested the parser on ONNX models converted from tensorflow as well as the ACAS Xu prototype networks in ONNX format found in the VNNLIB repository. A new python script in examples shows how to load and solve queries for ONNX format networks that have been added to examples/networks.

I also fixed a minor bug in Marabou.read_tf to change inputName to inputNames to reflect the multi-input changes made in MarabouNetworkTF.


kjulian3:
Thanks Chris! There's lots of valid ways to define on ONNX network, so I tried to make the equation functions robust, although that makes things more complicated. I think in the future there needs to be some sort of testing framework for the input parsers to make sure we're robust to many network types, and to make sure that a bug fix to accommodate new networks does not break functionality for other networks. But hopefully this is a good starting point for the ONNX parser.

guykatzz:
Hi Kyle,
Sorry for the delay in reviewing. The code looks good to me!

PR #221, opened 2020-02-26
Add support for different branching heuristics

anwu1219:
This PR is needed both for DnC Relu Splitting (the CAV submission) and for experimenting with different branching heuristics within the SmtCore.  
1. Each plConstraint keeps a score that will be used to select the branching plConstraint.
2. Add a pickBranchPLConstraint() method in the SmtCore that returns the PL Constraint with the highest score when branching is required. 
3. A strategy to pick the earliest ReLU is implemented (2 more SAT and 5 more UNSAT on all ACAS benchmarks with a 30 minutes timeout). 

Future PR will include the polarity based branching heuristic, and relusplitting in the DnC module.


src/engine/Engine.h
```diff
@@ -127,6 +128,17 @@ class Engine : public IEngine, public SignalHandler::Signalable
...
+      Pick the piecewise linear constraint for branching
+    */
+    PiecewiseLinearConstraint *pickBranchPLConstraint();
+
+    /*
+      Update the scores of each candidate branching PL constraints
+    */
+    void updateScores();
+
+
```
guykatzz: remove extra empty line

src/engine/Engine.h
```diff
@@ -170,6 +182,11 @@ class Engine : public IEngine, public SignalHandler::Signalable
     */
     List<PiecewiseLinearConstraint *> _plConstraints;
 
+    /*
+      The orderred set of candidate PL constraints for branching
```
guykatzz: Typo: ordered

src/engine/IEngine.h
```diff
@@ -69,6 +69,14 @@ class IEngine
     */
     virtual void reset() = 0;
     virtual List<unsigned> getInputVariables() const = 0;
+
+    virtual void updateScores()=0;
```
guykatzz: spaces: () = 0;

src/engine/MaxConstraint.h
```diff
@@ -125,6 +125,7 @@ class MaxConstraint : public PiecewiseLinearConstraint
  private:
     unsigned _f;
     Set<unsigned> _elements;
+    double _score;
```
guykatzz: If you define _score in the parent class, I don't think you can/should define it in the children without. I think this will mask the parent definition.

src/engine/ReluConstraint.h
```diff
@@ -186,7 +187,7 @@ class ReluConstraint : public PiecewiseLinearConstraint
     PhaseStatus getDirection() const;
 
 private:
-    unsigned _b, _f;
+    unsigned _b, _f, _layer;
```
guykatzz: This is a bit problematic, or at least is a shift in paradigm. So far, a PL constraint has no notion of a DNN, and hence of a layer. This is also how we present things in the ReluPlex paper - it is a general tool for solving LRA queries with ReLUs. So far, the only place that contained the notion of a DNN was the NLR. 

Maybe lets have a setScore() function, and have the AcasParser or someone else invoke it for the ReLU constraints?

[There's another change of paradigm here, and that is that constraints rank their splitting priority as opposed to the SMT Core deciding for them, but I'm okay with that]

src/engine/SmtCore.cpp
```diff
@@ -389,6 +394,14 @@ PiecewiseLinearConstraint *SmtCore::chooseViolatedConstraintForFixing( List<Piec
     return candidate;
 }
 
+void SmtCore::pickBranchPLConstraint()
+{
+    if ( _needToSplit && ( !_constraintForSplitting ) )
```
guykatzz: remove extra ( ).

Also, so far our terminology has been "splitting", not "branching". Does branching have a different meaning?

src/engine/tests/MockEngine.h
```diff
@@ -142,6 +142,16 @@ class MockEngine : public IEngine
...
+
+    void updateScores()
+    {
+    }
+
+    PiecewiseLinearConstraint *pickBranchPLConstraint()
+    {
+        return NULL;
+    }
+
```
guykatzz: Remove extra line

src/engine/MaxConstraint.h
anwu1219: Removed _score from children classes.

src/engine/SmtCore.cpp
anwu1219: renamed to "split"

src/engine/ReluConstraint.h
anwu1219: I see. That makes sense. I added a setScore() method, and let the AcasParser invoke it if we use the EarliestReLU divide strategy.

ahmed-irfan:
@anwu1219 there are conflict with the current master

anwu1219:
> @anwu1219 there are conflict with the current master

Fixed now.

guykatzz:
LGTM

PR #222, opened 2020-02-26
Add Constraint Violation Threshold as an command line option

anwu1219:
Give users the ability to change the constraint violation threshold.
Usage:
`./Marabou [network] [property] --split-threshold=1`
On ACAS benchmarks (property 1-4) with a 30 minutes timeout, 
split-threshold=1 solves 9 more SAT and 5 more UNSAT than split-threshold=20.

### DIFF
--- a/src/configuration/OptionParser.cpp
+++ b/src/configuration/OptionParser.cpp
@@ -78,7 +78,7 @@ void OptionParser::initialize()
           "out statistics in the beginning and end, 2: print out statistics during solving." )
         ( "split-threshold",
           boost::program_options::value<int>( &((*_intOptions)[Options::SPLIT_THRESHOLD]) ),
-          "Threshold to split" )
+          "Max number of tries to repair a relu before splitting" )
         ( "timeout-factor",
           boost::program_options::value<float>( &((*_floatOptions)[Options::TIMEOUT_FACTOR]) ),
           "(DNC) The timeout factor" )
--- a/src/configuration/OptionParser.cpp
+++ b/src/configuration/OptionParser.cpp
@@ -76,6 +76,9 @@ void OptionParser::initialize()
           boost::program_options::value<int>( &((*_intOptions)[Options::VERBOSITY]) ),
           "Verbosity of engine::solve(). 0: does not print anything (for DnC), 1: print"
           "out statistics in the beginning and end, 2: print out statistics during solving." )
+        ( "split-threshold",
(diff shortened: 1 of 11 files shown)


src/configuration/OptionParser.cpp
```diff
@@ -76,6 +76,9 @@ void OptionParser::initialize()
           boost::program_options::value<int>( &((*_intOptions)[Options::VERBOSITY]) ),
           "Verbosity of engine::solve(). 0: does not print anything (for DnC), 1: print"
           "out statistics in the beginning and end, 2: print out statistics during solving." )
+        ( "split-threshold",
+          boost::program_options::value<int>( &((*_intOptions)[Options::SPLIT_THRESHOLD]) ),
+          "Threshold to split" )
```
ahmed-irfan: add more explanation

ahmed-irfan: perhaps "Max number of tries to fix a relu before splitting"

src/configuration/Options.cpp
```diff
@@ -53,6 +53,7 @@ void Options::initializeDefaultValues()
     _intOptions[INITIAL_TIMEOUT] = 5;
     _intOptions[VERBOSITY] = 2;
     _intOptions[TIMEOUT] = 0;
+    _boolOptions[SPLIT_THRESHOLD] = 20;
```
ahmed-irfan: why boolOption?

src/engine/Marabou.cpp
```diff
@@ -102,6 +102,12 @@ void Marabou::prepareInputQuery()
             printf( "Property: None\n" );
 
         printf( "\n" );
+
+        /*
+          Step 3: extract options
+        */
+        unsigned splitThreshold = Options::get()->getInt( Options::SPLIT_THRESHOLD );
```
ahmed-irfan: getInt can return negative value. Need to guard that case

guykatzz:
Other than Ahmed's comments, LGTM.

src/configuration/OptionParser.cpp
anwu1219: changed comments

src/configuration/Options.cpp
anwu1219: whoops. Changed to intOptions

src/engine/Marabou.cpp
anwu1219: guarded

PR #223, opened 2020-02-27
Replay treestates in DnC

anwu1219:
Add option "--restore-tree-states". When turned on, in the DnC mode, the sub-queries will inherit it's (timed out) parent's tree state instead of starting from scratch.

Compared with no restore tree states on 180 ACAS benchmarks with a 30 minutes timeout:
solve 2 more SAT and 1 more UNSAT. ~20% wall time reduction on commonly solved instances.

### DIFF
--- a/src/engine/DnCManager.h
+++ b/src/engine/DnCManager.h
@@ -51,7 +51,7 @@ public:
     /*
       Perform the Divide-and-conquer solving
     */
-    void solve( unsigned timeoutInSeconds );
+    void solve( unsigned timeoutInSeconds, bool restoreTreeStates=true );
 
     /*
       Return the DnCExitCode of the DnCManager
@@ -81,7 +81,8 @@ private:
                           std::atomic_uint &numUnsolvedSubQueries,
                           std::atomic_bool &shouldQuitSolving,
                           unsigned threadId, unsigned onlineDivides,
-                          float timeoutFactor, DivideStrategy divideStrategy );
+                          float timeoutFactor, DivideStrategy divideStrategy,
+                          bool restoreTreeStates );
 
     /*
       Create the base engine from the network and property files,
--- a/src/engine/DnCWorker.cpp
+++ b/src/engine/DnCWorker.cpp
@@ -63,7 +63,6 @@ void DnCWorker::setQueryDivider( DivideStrategy divideStrategy )
 
 void DnCWorker::popOneSubQueryAndSolve( bool restoreTreeStates )
 {
-    std::cout << restoreTreeStates << std::endl;
     SubQuery *subQuery = NULL;
     // Boost queue stores the next element into the passed-in pointer
     // and returns true if the pop is successful (aka, the queue is not empty
@@ -73,6 +72,8 @@ void DnCWorker::popOneSubQueryAndSolve( bool restoreTreeStates )
         String queryId = subQuery->_queryId;
         auto split = std::move( subQuery->_split );
         std::unique_ptr<SmtState> smtState = nullptr;
+        if ( restoreTreeStates && subQuery->_smtState )
+            smtState = std::move( subQuery->_smtState );
         unsigned timeoutInSeconds = subQuery->_timeoutInSeconds;
 
         // Reset the engine state
@@ -88,9 +89,11 @@ void DnCWorker::popOneSubQueryAndSolve( bool restoreTreeStates )
 
         bool fullSolveNeeded = true; // denotes whether we need to solve the subquery
         if ( restoreTreeStates && smtState )
+        {
+            std::cout << "Restoring tree states" << std::endl;
             fullSolveNeeded = _engine->restoreSmtState( *smtState );
-
-        IEngine::ExitCode result;
+        }
+        IEngine::ExitCode result = IEngine::NOT_DONE;
         if ( fullSolveNeeded )
         {
             _engine->solve( timeoutInSeconds );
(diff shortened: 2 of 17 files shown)


src/engine/DnCWorker.h
```diff
@@ -36,7 +36,7 @@ class DnCWorker
       Pop one subQuery, solve it and handle the result
       Return true if the DnCWorker should continue running
     */
-    void popOneSubQueryAndSolve();
+    void popOneSubQueryAndSolve( bool restoreTreeStates=true );
```
guykatzz: spaces: restoreTreeStates = true

src/engine/SmtCore.cpp
```diff
@@ -397,6 +397,58 @@ PiecewiseLinearConstraint *SmtCore::chooseViolatedConstraintForFixing( List<Piec
...
+
+void SmtCore::storeSmtState(SmtState & smtState)
+{
+    smtState._impliedValidSplitsAtRoot = _impliedValidSplitsAtRoot;
+
+    for ( auto &stackEntry : _stack )
+        smtState._stack.append( duplicateStackEntry( *stackEntry ) );
+}
+
+StackEntry * SmtCore::duplicateStackEntry(const StackEntry & stackEntry)
```
guykatzz: I think this should be a method of StackEntry, since it seems to have nothing to do with SmtCore.
Also, the spacing is off:
StackEntry *SmtCore::duplicateStackEntry( const StackEntry &stackEntry )

src/engine/IEngine.h
```diff
@@ -53,6 +54,18 @@ class IEngine
     virtual void restoreState( const EngineState &state ) = 0;
     virtual void setNumPlConstraintsDisabledByValidSplits( unsigned numConstraints ) = 0;
 
+    /*
+      Apply the stack to the newly created SmtCore, returns false if UNSAT is
+      found in this process.
+    */
+
+    virtual void storeSmtState( SmtState &smtState ) = 0;
+
```
guykatzz: Remove empty line

src/engine/Engine.cpp
```diff
@@ -1934,6 +1934,64 @@ void Engine::updateDirections()
                 constraint->updateDirection();
 }
 
+bool Engine::restoreSmtState( SmtState & smtState )
```
guykatzz: SmtState &smtState

src/engine/SmtCore.h
```diff
@@ -98,6 +100,16 @@ class SmtCore
...
 
+    /*
+      Replay a stackEntry
+    */
+    void replayStackEntry( StackEntry *stackEntry );
+
+    /*
+      Store the current stack into smtState
+    */
+    void storeSmtState( SmtState &smtState );
```
guykatzz: The lack of symmetry here is unclear. Why not have "store" and "restore" methods? Why do you want to replay each stack entry individually? That would be useful only if you wanted to replay some but not others - is this the case? Right now, the logic is at the Engine instead of the SmtCore.

src/engine/Engine.cpp
```diff
@@ -1934,6 +1934,64 @@ void Engine::updateDirections()
...
+{
+    try
+    {
+        // Step 1: all implied valid splits at root
```
guykatzz: See my other comment - consider moving this logic to SmtCore

src/engine/DnCWorker.cpp
```diff
@@ -83,9 +86,21 @@ void DnCWorker::popOneSubQueryAndSolve()
 
         // Apply the split and solve
         _engine->applySplit( *split );
-        _engine->solve( timeoutInSeconds );
 
-        IEngine::ExitCode result = _engine->getExitCode();
+        bool fullSolveNeeded = true; // denotes whether we need to solve the subquery
```
guykatzz: Name is confusing. Maybe call this "queryIsUnsat" or something.

src/engine/DnCManager.h
```diff
@@ -51,7 +51,7 @@ class DnCManager
     /*
       Perform the Divide-and-conquer solving
     */
-    void solve( unsigned timeoutInSeconds );
+    void solve( unsigned timeoutInSeconds, bool restoreTreeStates=true );
```
ahmed-irfan: perhaps, it is better to be aligned with the default option for restoreTresStates.

So, here it should be `bool restoreTreeStates=false`

src/engine/DnCWorker.cpp
```diff
@@ -83,9 +86,21 @@ void DnCWorker::popOneSubQueryAndSolve()
...
+        if ( restoreTreeStates && smtState )
+            fullSolveNeeded = _engine->restoreSmtState( *smtState );
+        IEngine::ExitCode result = IEngine::NOT_DONE;
+        if ( fullSolveNeeded )
+        {
+            _engine->solve( timeoutInSeconds );
+            result = _engine->getExitCode();
+        } else
```
ahmed-irfan: format

src/engine/DnCWorker.h
ahmed-irfan: align it with the default option

ahmed-irfan: also,
align the `restoreTreeStates` with the default option

ahmed-irfan:
there are conflicts

src/engine/StackEntry.h
```diff
@@ -0,0 +1,45 @@
...
+/*
+  A stack entry consists of the engine state before the split,
+  the active split, the alternative splits (in case of backtrack),
+  and also any implied splits that were discovered subsequently.
+*/
+struct StackEntry
+{
+public:
+    PiecewiseLinearCaseSplit _activeSplit;
+    List<PiecewiseLinearCaseSplit> _impliedValidSplits;
```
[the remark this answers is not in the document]
ahmed-irfan: i like `impliedLemmas`

anwu1219:
Close since we would re-implement this after refactoring.

PR #224, opened 2020-02-28
perform sbt on both branches

anwu1219:
Right now, we only perform SBT on the first branch of a PL constraint. This PR address that.
Comparison on ACAS benchmarks using DnC mode (8 core)
SBT on both branches solves 1 more SAT, 5 more UNSAT than the current version.


guykatzz:
LGTM

PR #228, opened 2020-03-24
Absolute value PL constraint

guykatzz:


src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -0,0 +1,454 @@
...
+    _assignment[variable] = value;
+}
+
+/*
+  The variable watcher notification callbacks, about a change in a variable's value or bounds.
+  suppose A < x_b < B, C < x_f < D
+  if variable == x_b then:
+    A < 0 & B < 0 then: max{|B|, C} < x_f < min{|A|, D}
+    A < 0 & B > 0 then: max{0, C} < x_f < min{max{|A|, B}, D}
+    A > 0 & B < 0 then: ------
```
AleksandarZeljic: Spell out that  -----  means UNSAT/Infeasable/Conflict

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -0,0 +1,454 @@
...
+        if ( variable == _b )
+        {
+            if ( bound < 0 )
+            {
+                double fUpperBound = FloatUtils::max( -bound, _upperBounds[_b] );
+                _constraintBoundTightener->registerTighterUpperBound( _f, fUpperBound );
+            }
+            else
+            {
+                // Phase is fixed, don't care about this case
```
AleksandarZeljic: Is this always going to be the case? From the perspective of generating explanations in the future.

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -0,0 +1,454 @@
...
+        else
+        {
+            // F's lower bound can only cause bound propagations if it
+            // fixes the phase of the constraint, so no need to bother
+        }
+
+        // Also, if for some reason we only know a negative lower bound for f,
+        // we attempt to tighten it to 0
+        if ( bound < 0 && variable == _f )
+            _constraintBoundTightener->registerTighterLowerBound( _f, 0 );
```
AleksandarZeljic: This should be done only once, ideally at the start of solving. What is the reason to keep attempting it?

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -0,0 +1,454 @@
...
+/*
+  The variable watcher notification callbacks, about a change in a variable's value or bounds.
+  suppose A < x_b < B, C < x_f < D
+  if variable == x_b then:
+    A < 0 & B < 0 then: max{|B|, C} < x_f < min{|A|, D}
+    A < 0 & B > 0 then: max{0, C} < x_f < min{max{|A|, B}, D}
+    A > 0 & B < 0 then: ------
+    A > 0 & B > 0 then: max{A, C} < x_f < min{B, D}
+  if variable == x_f then:
+    C > 0 & D > 0 then: min{max{-D, A}, max{A, C}} < x_b < max{min{-C, A}, min{B, D}}
```
AleksandarZeljic: Add breakdown by cases for this line, to match what the code is doing, i.e. no updates from C to A and B, and updates from D to A and B.

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+List<PiecewiseLinearConstraint::Fix> AbsoluteValueConstraint::getSmartFixes( ITableau */* tableau */ ) const
+{
+    return getPossibleFixes();
+}
+
+List<PiecewiseLinearCaseSplit> AbsoluteValueConstraint::getCaseSplits() const
+{
+    if ( _phaseStatus != PhaseStatus::PHASE_NOT_FIXED )
+        throw MarabouError( MarabouError::REQUESTED_CASE_SPLITS_FROM_FIXED_CONSTRAINT );
```
AleksandarZeljic: Is this an error we should to recover from? Should we just assert this as an invariant?

AleksandarZeljic: Line 268: uses the assert form for the same check.

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -0,0 +1,454 @@
...
+{
+    ASSERT( _phaseStatus != PhaseStatus::PHASE_NOT_FIXED );
+
+    if ( _phaseStatus == PhaseStatus::PHASE_POSITIVE )
+        return getPositiveSplit();
+
+    return getNegativeSplit();
+}
+
+void AbsoluteValueConstraint::eliminateVariable( unsigned /* variable */, double /* fixedValue */ )
```
AleksandarZeljic: A general question, how can we eliminate variables?

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -0,0 +1,454 @@
...
+            _upperBounds.exists( _b ) && _upperBounds.exists( _f ) );
+
+    // Upper bounds
+    double bUpperBound = _upperBounds[_b];
+    double fUpperBound = _upperBounds[_f];
+    // Lower bounds
+    double bLowerBound = _lowerBounds[_b];
+    double fLowerBound = _lowerBounds[_f];
+
+    // F's lower bound should always be non-negative
```
AleksandarZeljic: Second place where we return this tightening, I believe.

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -0,0 +1,454 @@
...
+    {
+        // Phase undetermined, b can be either positive or negative, f strictly positive
+        tightenings.append( Tightening( _b, -fUpperBound, Tightening::LB ) );
+        tightenings.append( Tightening( _b, fUpperBound, Tightening::UB ) );
+        tightenings.append( Tightening( _f, FloatUtils::max( -bLowerBound, bUpperBound ), Tightening::UB ) );
+
+        // Below we test if the phase has actually become fixed
+        if ( fLowerBound > -bLowerBound )
+        {
+            // Positive phsae
```
AleksandarZeljic: Typo for phase, below also

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    // Option 2: b's range is strictly negative:
+    if ( _upperBounds.exists( _b ) && _upperBounds[_b] <= 0 )
+    {
+        setPhaseStatus( PHASE_NEGATIVE );
+        return;
+    }
+
+    if ( !_lowerBounds.exists( _f ) )
```
AleksandarZeljic: What is this case?

src/engine/AbsoluteValueConstraint.h
```diff
@@ -0,0 +1,171 @@
...
+      Currently not implemented, just calls getPossibleFixes().
+    */
+    List<PiecewiseLinearConstraint::Fix> getSmartFixes( ITableau *tableau ) const;
+
+    /*
+      Returns the list of case splits that this piecewise linear
+      constraint breaks into. These splits need to complementary,
+      i.e. if the list is {l1, l2, ..., ln-1, ln},
+      then ~l1 /\ ~l2 /\ ... /\ ~ln-1 --> ln.
+    */
```
AleksandarZeljic: Update the comment for this abs constraint in particular?

src/engine/AbsoluteValueConstraint.h
```diff
@@ -0,0 +1,171 @@
...
+    /*
+      Preprocessing related functions, to inform that a variable has
+      been eliminated completely because it was fixed to some value,
+      or that a variable's index has changed (e.g., x4 is now called
+      x2). constraintObsolete() returns true iff and the constraint
+      has become obsolote as a result of variable eliminations.
+     */
+    void eliminateVariable( unsigned variable, double fixedValue );
+    void updateVariableIndex( unsigned oldIndex, unsigned newIndex );
+    bool constraintObsolete() const;
```
AleksandarZeljic: Maybe name the bool returning test functions with is prefix, i.e. isConstraintObsolete()? It would distinguish from imperative function names and make the code easier to read.

src/engine/Engine.cpp
```diff
@@ -1125,6 +1125,8 @@ void Engine::extractSolution( InputQuery &inputQuery )
             if ( _preprocessor.variableIsFixed( variable ) )
             {
                 inputQuery.setSolutionValue( i, _preprocessor.getFixedValue( variable ) );
+                inputQuery.setLowerBound( i, _preprocessor.getFixedValue( variable ) );
```
AleksandarZeljic: Why are these calls added now? Are they related to abs constraint?

src/engine/AbsoluteValueConstraint.cpp
guykatzz: The code comment was stale, removing it.

src/engine/AbsoluteValueConstraint.cpp
guykatzz: Yes, this is here only for understandability. If the phase has become fixed (elsewhere), that the explanations should be generated elsewhere, also.

src/engine/AbsoluteValueConstraint.cpp
guykatzz: Stale comment, removed.

src/engine/AbsoluteValueConstraint.cpp
guykatzz: The preprocessor eliminates variables if they become fixed (5 <= x <= 5), or if identical varaibles are discovered (x=y; one is eliminated from all other equations).

src/engine/AbsoluteValueConstraint.cpp
guykatzz: Initially, when the PL constraint is just created, it doesn't know any bounds yet - this is just to be on the safe side.

src/engine/AbsoluteValueConstraint.h
guykatzz: Currently this convention is not really followed, not even in this particular class. Is it in our coding guidelines?

src/engine/Engine.cpp
guykatzz: These are small issues that were discovered as part of the testing of the Abs Constraint

src/engine/AbsoluteValueConstraint.cpp
guykatzz: Indeed, this should only apply in the beginning, but we don't know whether it's been done before or not. I refactored it a bit, for efficiency (moved it into the previous if condition).

src/engine/AbsoluteValueConstraint.cpp
guykatzz: Fixed

src/engine/AbsoluteValueConstraint.cpp
guykatzz: Yes, there are two mechanisms for getting tighter bounds: through the "notify" functions, where the constraint initiates sending the information; or through the getEntailedTightengings function, where the caller initiates the process.

Issue #226, opened 2020-03-22
When running `Marabou.cpp` - asserts arn't called even with debug mode 


Yoni-Mantzur:
When running the main module - `Marabou.cpp` on debug mode, assertions are not called. Probably due to missing `--DDEBUG` flag

yuvaljacoby:
Are you sure clion is adding DEBUG flag? I know that usually NDEBUG is used to indicate release mode (anyway that is what cmake is doing).
I just pushed [use_ndebug](https://github.com/NeuralNetworkVerification/Marabou/tree/use_ndebug) branch,can you try it out? if it doesn't t work take a look at src/common/Debug.h or late me know

Yoni-Mantzur:
@yuvaljacoby the fix looks good, tested on `debug` mode and `release` mode.

PR #229, opened 2020-03-24
Use ndebug

yuvaljacoby:
Solves issue: #226

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,10 +104,10 @@ endif()
 add_definitions(-DRESOURCES_DIR="${RESOURCES_DIR}")
 
 if (NOT MSVC)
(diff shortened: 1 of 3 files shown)


guykatzz:
So how do we trigger debug mode now? This NDEBUG flag is not explicit in the CMakeLists file, right?

yuvaljacoby:
CMAKE be default adds the NDEBUG flag
To compile in debug mode we need to ask that from cmake, you do as follows: 

    cmake .. -DCMAKE_BUILD_TYPE=Debug
    cmake --build .

that is pretty standard with cmake, but I can emphasize in the README if needed

guykatzz:
Okay. Please add a comment to the Readme, and then you can merge AFAIC.

PR #232, opened 2020-03-28
Improved Relu splits

guykatzz:
In a ReLU's active split, if we already have an auxiliary variable aux = f - b, we don't add a new equation; instead, we make aux strictly 0.

### DIFF
--- a/src/engine/ReluConstraint.cpp
+++ b/src/engine/ReluConstraint.cpp
@@ -491,7 +491,7 @@ PiecewiseLinearCaseSplit ReluConstraint::getActiveSplit() const
     if ( _auxVarInUse )
     {
         // Special case: aux var in use.
-        // Because aux = f - b, we just set aux = 0.
+        // Because aux = f - b and aux >= 0, we just add that aux <= 0.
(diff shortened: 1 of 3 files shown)


src/engine/ReluConstraint.cpp
```diff
@@ -487,11 +487,22 @@ PiecewiseLinearCaseSplit ReluConstraint::getActiveSplit() const
...
-    activeEquation.addAddend( 1, _b );
-    activeEquation.addAddend( -1, _f );
-    activeEquation.setScalar( 0 );
-    activePhase.addEquation( activeEquation );
+
+    if ( _auxVarInUse )
+    {
+        // Special case: aux var in use.
+        // Because aux = f - b, we just set aux = 0.
+        activePhase.storeBoundTightening( Tightening( _aux, 0.0, Tightening::UB ) );
```
ahmed-irfan: perhaps, it is worth mentioning that we have already added the bound `_aux >= 0`

guykatzz: Done

PR #236, opened 2020-03-30
Rm python2

ahmed-irfan:
Python2 reached its end-of-life in Jan 2020.

This PR drops support for python2 in CMake


yuvaljacoby:
as far as I remember I added it because it didn't work on some cluster or server with old versions.
It's basically there to try to avoid frustration.
Maybe we can keep at until something break and then just remove? for now it doesn't hurt...

PR #237, opened 2020-03-30
use sat/unsat instead of SAT/UNSAT

ahmed-irfan:
This PR changes the result output printing. With this change, we can use the cvc4-cluster-tools for running experiments and analyzing their results


guykatzz:
Hi Ahmed,
The Travis tests seem to be failing for this PR.

ahmed-irfan:
looking into it

yuvaljacoby:
I think it's the DnCManager 

I would also suggest having sat/unsat as a define somewhere in common directory and import it everywhere, so next time we need to change it will be easier :)

anwu1219:
As Yuval suggested, I think this method in DnCManager.cpp needs also be updated.
https://github.com/NeuralNetworkVerification/Marabou/blob/38d97ab7c2beb51e5ab0a7e65318cee0d54f22dc/src/engine/DnCManager.cpp#L248

anwu1219:
LGTM

Issue #239, opened 2020-04-10
transA and transB support for GEMM


samuelemarro:
ONNX's GEMM operation supports the optional parameters `transA` and `transB`, which respectively transpose the input and weight tensors. Unfortunately, maraboupy assumes that both of these parameters are false. This causes an assertion error, since if a tensor is transposed, the dimensions will be reversed:
https://github.com/NeuralNetworkVerification/Marabou/blob/41ccc59a1bcadd99b8038317e25cdba34a017677/maraboupy/MarabouNetworkONNX.py#L490

Since PyTorch converts `nn.Linear` modules into GEMM operations with `transB=1`, many exported PyTorch models cannot be opened by Marabou.

kjulian3:
Thanks for pointing this out. The test networks I created didn't use those options, so they didn't get implemented. There may be other ONNX options or functionality that hasn't been supported yet, so please let us know if you run into other issues so we can make improvements. I'll work on a pr for transA and transB.

samuelemarro:
Thank you.

kjulian3:
@samuelemarro I submitted pull request #242. Can you take a look and let me know if this solves your issue?

samuelemarro:
Tested with a simple FeedForward network and some dummy inputs. It converts the model without errors and Marabou's output matches both PyTorch's and ONNX's output. Thank you for your help.

PR #242, opened 2020-04-14
Onnx gemm attributes

kjulian3:
This pull request addresses #239 by adding support for the four Gemm attributes (transA, transB, alpha, and beta), as well as fixing an edge case for the Matmul function. These bugs were found using networks exported from PyTorch.


maraboupy/MarabouNetworkONNX.py
```diff
@@ -522,7 +540,7 @@ def matMulEquations(self, node, makeEquations):
         shape2 = self.shapeMap[inputName2]
         assert shape1[-1] == shape2[0]
         self.shapeMap[nodeName] = shape1[:-1] + shape2[1:]
-        
+
```
anwu1219, line 543: remove extra space

kjulian3: Removed space

PR #245, opened 2020-04-20
Basic bound tightening capabilities for the NLR

guykatzz:
(not currently called by the Engine)

Support for standard, interval arithmetic bound propagation.


chelseas:
Curious, how is this different from the bound tightening that existed before?

guykatzz:
Most of our bound tightening is done locally: we look at a single constraint, or a single equation, or the current basis matrix, and use them to do some bound tightening. The NetworkLevelReasoner is intended to run more complex tightening procedures, that use the network structure. In this PR, I added support for doing layer-by-layer interval propagation. This is needed for some work that we're doing on supporting Sigmoids. In the next PR I plan to merge the Symbolic Bound Tightening procedure into the network level reasoner, as well.

src/engine/NetworkLevelReasoner.cpp
```diff
@@ -330,6 +417,108 @@ void NetworkLevelReasoner::updateVariableIndices( const Map<unsigned, unsigned>
...
+
+                switch ( _neuronToActivationFunction[index] )
+                {
+                case ReLU:
+                    _lowerBoundsActivations[i][j] = lb > 0 ? lb : 0;
+                    _upperBoundsActivations[i][j] = ub;
+                    break;
+
+                default:
+                    ASSERT( false );
```
ahmed-irfan, line 481: is NLR going to be used only for networks with Relus?

guykatzz:
The plan is to add support for additional kinds of networks - hopefully in
PR #3 :).

src/engine/NetworkLevelReasoner.cpp
ahmed-irfan: OK. So, how about if we don't have the assertion and then the code won't do anything on other networks. Right?

guykatzz: It might produce erroneous results, if it doesn't take into account the effects of other functions. For example, if we have absolute value and we let a negative value through.

ahmed-irfan: i see. hmm...

I am worried that in the production mode, we will not detect this. Perhaps, we should throw an exception here.

guykatzz: Good point. Fixed.

PR #248, opened 2020-04-22
Preprocess the input query only once in DnC

anwu1219:
Currently in DnC, inputQuery is preprocessed multiple times. 
This PR preprocess the inputQuery once with the baseEngine in DnCManager, and the worker engines only copy over the preprocessed input query.

### DIFF
--- a/src/engine/DnCManager.cpp
+++ b/src/engine/DnCManager.cpp
@@ -32,7 +32,7 @@
 #include <thread>
 
 void DnCManager::dncSolve( WorkerQueue *workload, std::shared_ptr<Engine> engine,
-                           InputQuery *inputQuery,
+                           InputQuery &inputQuery,
                            std::atomic_uint &numUnsolvedSubQueries,
                            std::atomic_bool &shouldQuitSolving,
                            unsigned threadId, unsigned onlineDivides,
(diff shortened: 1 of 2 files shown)


src/engine/DnCManager.cpp
```diff
@@ -30,8 +30,12 @@
 #include <chrono>
 #include <cmath>
 #include <thread>
+#include <fstream>
+
+#include "util/util.hh"
```
ahmed-irfan, line 35: where is this file in our repo?

is the filename correct? (`*.hh`)

anwu1219: ah, they shoudn't be there. I removed those.

src/engine/DnCManager.cpp
```diff
@@ -138,8 +141,11 @@ void DnCManager::solve( unsigned timeoutInSeconds )
     std::list<std::thread> threads;
     for ( unsigned threadId = 0; threadId < _numWorkers; ++threadId )
     {
-        threads.push_back( std::thread( dncSolve, workload,
-                                        _engines[ threadId ],
+        InputQuery *inputQuery = new InputQuery();
```
ahmed-irfan, line 144: who is going to deallocate the memory?

anwu1219: Thanks for pointing this out. I wrap the InputQuery with a unique ptr. Since there will be a copy of input query for each worker, it still makes sense to allocate them on the heap.

PR #251, opened 2020-04-25
Symbolic bound tightening over absolute values

guykatzz:
Teach the network level reasoner to work with AbsoluteValue constraints, and not just ReLUs. One motivating example is to have better support for adversarial robustness queries that use L1 norm. 

This is PR #3 in this series


src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -259,8 +259,10 @@ PiecewiseLinearCaseSplit AbsoluteValueConstraint::getValidCaseSplit() const
     return getNegativeSplit();
 }
 
-void AbsoluteValueConstraint::eliminateVariable( unsigned /* variable */, double /* fixedValue */ )
+void AbsoluteValueConstraint::eliminateVariable( unsigned variable, double /* fixedValue */ )
```
ahmed-irfan, line 262: what is the second argument for?

ahmed-irfan:
there are conflicts

src/engine/AbsoluteValueConstraint.cpp
guykatzz: This is an interface that all PLConstraints implement - it tells you that a variable has been eliminated (because it became constant), and the value it has been assigned.

The motivation for including the second argument was for MaxConstraint. If y = max(x1,x2,x3), and x3 has been eliminated, you care about the value - because now y can never go below that value. 

As I was writing the reply, I noticed that this was actually not properly done in MaxConstraint, so I'll open an issue :).

guykatzz:
Fixed the conflicts

src/engine/AbsoluteValueConstraint.cpp
ahmed-irfan: I see. Thanks for explaining this.

PR #253, opened 2020-04-27
Auto construction of a network level reasoner

guykatzz:
When a user has not provided a network level reasoner as part of the input query, Marabou will attempt to automatically figure out the topology of the network and construct one itself. This is needed especially for symbolic bound tightening.


src/engine/NetworkLevelReasoner.cpp
```diff
@@ -1151,6 +1155,93 @@ void NetworkLevelReasoner::log( const String &message )
...
+                for ( unsigned k = 0; k < _layerSizes[i-1]; ++k )
+                {
+                    double weight = _weights[i-1][k * layerSize + j];
+                    if ( FloatUtils::isZero( weight ) )
+                        continue;
+
+                    if ( weight < 0 )
+                        printf( " - %.5lfx%u", -weight, _indexToActivationResultVariable[Index( i - 1, k )] );
+                    else
+                        printf( " + %.5lfx%u", weight, _indexToActivationResultVariable[Index( i - 1, k )] );
```
AleksandarZeljic, line 1230: If you do not care for the space between the sign and the value, you can add the + flag in the formatting string " %+.5lfx%u"  to print out the sign for both positive and negative numbers, which would eliminate the need for the if statement.

src/engine/Preprocessor.cpp
```diff
@@ -781,6 +794,278 @@ void Preprocessor::dumpAllBounds( const String &message )
...
+    List<unsigned> currentLayer;
+    List<unsigned> outputLayer;
+    unsigned currentLayerIndex = 1;
+
+    // The input and output layer should be defined in the input query
+    inputLayer = _preprocessed.getInputVariables();
+    previousLayer = inputLayer;
+    unsigned numVariablesHandledSoFar = inputLayer.size();
+    outputLayer = _preprocessed.getOutputVariables();
+    layerToLayerSize[0] = inputLayer.size();
```
AleksandarZeljic, line 874: Maybe move this to line above 873 since it relates to input layer (there is a line referring to the output layer in between). If moved above 872, 872 can be rephrased to use layerToLayerSize[0], which might be more readable since it will likely be used later.

src/engine/Preprocessor.cpp
```diff
@@ -781,6 +794,278 @@ void Preprocessor::dumpAllBounds( const String &message )
...
+
+    // Store the input neurons in the topology
+    unsigned count = 0;
+    for ( const auto &neuron : inputLayer )
```
AleksandarZeljic, line 878: Activation function will be unitialized, maybe add Identity to PiecewiseLinearFunctionType to denote input layer nodes? (in case we ever need to distinguish nodes on the fly, without the topology)

src/engine/Preprocessor.cpp
```diff
@@ -781,6 +794,278 @@ void Preprocessor::dumpAllBounds( const String &message )
...
+            if ( eq._type != Equation::EQ )
+                continue;
+
+            Set<unsigned> eqVars = eq.getParticipatingVariables();
+            for ( const auto &var : previousLayer )
+                eqVars.erase( var );
+
+            if ( eqVars.size() == 1 )
+            {
+                // We have identified the weithed sum for a new neuron
```
AleksandarZeljic, line 904: Typo in the comment - weighted

src/engine/Preprocessor.cpp
```diff
@@ -781,6 +794,278 @@ void Preprocessor::dumpAllBounds( const String &message )
...
+    {
+        currentLayer.clear();
+        for ( const auto &eq : equations )
+        {
+            // Only consider equalities
+            if ( eq._type != Equation::EQ )
+                continue;
+
+            Set<unsigned> eqVars = eq.getParticipatingVariables();
+            for ( const auto &var : previousLayer )
```
AleksandarZeljic, line 899: In case of variables skipping layers, even though less efficient, we could go through all the already processed variables? There is the issue of indexing, but least layer-depth indexing is probably a reasonable normal form for such nodes.

src/engine/Preprocessor.cpp
```diff
@@ -781,6 +794,278 @@ void Preprocessor::dumpAllBounds( const String &message )
...
+    while ( true )
+    {
+        currentLayer.clear();
+        for ( const auto &eq : equations )
+        {
+            // Only consider equalities
+            if ( eq._type != Equation::EQ )
+                continue;
+
+            Set<unsigned> eqVars = eq.getParticipatingVariables();
```
AleksandarZeljic, line 898: Storing this set across iterations might be more efficient, than repeatedly attempting to reduce it to one. Even though this will probably be a rare case. 

This would also address my next comment because it would be sufficient to eliminate just processed variables since the effects would accumulate across layer iterations.

src/engine/Preprocessor.cpp
```diff
@@ -781,6 +794,278 @@ void Preprocessor::dumpAllBounds( const String &message )
...
+                // We have identified the weithed sum for a new neuron
+                NeuronInformation neuronInfo;
+                neuronInfo._weightedSumVariable = *eqVars.begin();
+                neuronInfo._weightedSumEquation = &eq;
+
+                Index index( currentLayerIndex, currentLayer.size() );
+                indexToNeuron[index] = neuronInfo;
+                currentLayer.append( *eqVars.begin() );
+
+                variableToIndex[*eqVars.begin()] = index;
```
AleksandarZeljic, line 913: Maybe it makes sense from readability perspective to name *eqVars.begin(), or to use neuronInfo._weightedSumVariable in subsequent calls?

src/engine/Preprocessor.cpp
```diff
@@ -781,6 +794,278 @@ void Preprocessor::dumpAllBounds( const String &message )
...
+            }
```
AleksandarZeljic, line 914: After adding a weighted sum variable from an equation we can remove it from the set? To avoid needless iterations in the future?

src/engine/Preprocessor.cpp
```diff
@@ -781,6 +794,278 @@ void Preprocessor::dumpAllBounds( const String &message )
...
+        // If a new layer has not been discovered, we're done
+        if ( currentLayer.empty() )
+            break;
+
+        numVariablesHandledSoFar += currentLayer.size();
+        layerToLayerSize[currentLayerIndex] = currentLayer.size();
+
+        // If this layer is the output layer, we're done
+        bool outputLayerFound = false;
+        if ( currentLayer.size() == outputLayer.size() )
```
AleksandarZeljic, line 926: Is this a safe condition to identify output layer?

src/engine/Preprocessor.cpp
```diff
@@ -781,6 +794,278 @@ void Preprocessor::dumpAllBounds( const String &message )
...
+        NeuronInformation neuronInfo;
+        neuronInfo._activationVariable = neuron;
+        Index index( 0, count );
+        indexToNeuron[index] = neuronInfo;
+        variableToIndex[neuron] = index;
+        ++count;
+    }
+
+    // Now, attempt to figure out the topology, layer by layer
+    while ( true )
```
AleksandarZeljic, line 889: Should we use the remaining number of nodes to identify in the loop condition? If we cannot identify new layers and we have non-zero remaining nodes we can report a reconstruction failure. Otherwise we can use that number to identify the output layer more reliably (i.e. when remainingNodes == outputLayer.size())...

src/engine/Preprocessor.cpp
guykatzz: Done

src/engine/NetworkLevelReasoner.cpp
guykatzz: Done

src/engine/Preprocessor.cpp
guykatzz: Done

src/engine/Preprocessor.cpp
guykatzz: This is currently compatible with the NLR that is constructed through the ACAS parser. I agree that we could add the identity function, but will leave this for the future.

src/engine/Preprocessor.cpp
guykatzz: I agree that layer skipping is a nice feature. We should probably add it in the future. However, currently the SBT mechanism would break if we did that, so this is a topic for a separate PR.

src/engine/Preprocessor.cpp
guykatzz: Done

src/engine/Preprocessor.cpp
guykatzz: Done

guykatzz: This is a little tricky. This list of equations is a reference to the original equations. In order to accomplish this we would need to work on a copy of the list, which might defeat the purpose.

src/engine/Preprocessor.cpp
guykatzz: Fixed

src/engine/Preprocessor.cpp
guykatzz: I didn't understand this comment. What does it mean about the topology if an equation's variables appear across multiple layers?

src/engine/Preprocessor.cpp
guykatzz: I think cycling through the variables is slightly better, because it accounts for a common scenario where there are additional variables that are not included in the topology. For example, some delta variable that is used for an adversarial robustness bound.

src/engine/Preprocessor.cpp
guykatzz: There might be nodes not included in the topology, see above.

guykatzz:
Hi Aleks,
Could you please have a look at the changes?

src/engine/Preprocessor.cpp
AleksandarZeljic: Maybe I made an assumption that's incorrect:

I assumed that the equation is a weighted sum of inputs = backwards-facing-variable, so if some variables skip levels then there is no guarantee that all members of the weighted sum belong to the same level, so they may be discharged at different times. So what I was suggesting was a mapping {vars} -> equation, and through iterations we eliminate variables from the set, and when the variable set is empty we know the position of the activation function.

src/engine/Preprocessor.cpp
AleksandarZeljic: I had the same assumption/idea as in comment on line 895-898. I see what you mean though.

src/engine/Preprocessor.cpp
AleksandarZeljic: Ah, I see.

src/engine/NetworkLevelReasoner.cpp
```diff
@@ -1166,6 +1170,90 @@ void NetworkLevelReasoner::log( const String &message )
...
+                        continue;
+
+                    printf( " %+.5lfx%u", -weight, _indexToActivationResultVariable[Index( i - 1, k )] );
+                }
+
+                double bias = _bias[Index( i, j )];
+                if ( bias > 0 )
+                    printf( " + %.2lf", bias );
+                else
+                    printf( " - %.2lf", -bias );
```
AleksandarZeljic, line 1249: This can also be compacted by moving sign into the formatter.

Issue #244, opened 2020-04-15
Testing framework for Maraboupy


kjulian3:
We don't have a formal testing framework yet for maraboupy and the input parsers. I have a set of test networks I've been using to validate the tensorflow and onnx parsers, but incorporating these networks into automatic tests would tell us what parts of the code are still untested and ensure that future code changes don't introduce new bugs.

I can work on setting up the framework, but I'm not sure the best way to do that. Does anyone have any recommendations for setting up the testing framework, or is someone else currently working on this?

yuvaljacoby:
I agree this is important! 
Not only to make sure code does not break, it's also a great entry point for new users.

For framework I am usually working with pytest
Can you start and write some tests? 
I can later add them to the build process and travis

PR #254, opened 2020-04-28
Onnx tests

kjulian3:
First step in setting up a testing framework for python parsers to address #244. This pr adds a simple ONNX test case and adds the test to travis.


kjulian3:
I've added tests to travis using the pytest package, and the tests run for all three build types. Since the language is cpp, I had to manually add python packages pip and setuptools, and update those packages. The windows build puts the compiled MarabouCore into the Release subfolder, so that has to be copied up a level to maraboupy to work correctly. The pr shows one example test case for a fully-connected, ReLU-activated network defined with ONNX. More cases can be added by adding additional functions in test_onnx.py, and additional files in the maraboupy/test folder will be run as well, so we can write a similar test_tf.py as well. Future PRs will add more tests and expand testing to tensorflow.

kjulian3:
Lastly, pytest can also provide a simple code coverage report via pytest-cov. If we set it up with something like coveralls, we can get some nice interactive reports on what lines of the code have been tested, and which have yet to be tested. It would look something like this: https://coveralls.io/github/sisl/NNet

Are we interested in setting up coveralls, or some other form of code coverage?

yuvaljacoby:
Two small comments:
* Did you try to work with travis cache ? it can cache a folder between iterations, then we will not need to download python every time (I see it takes only ~30 seconds and only on windows so it is not a must). 
* Do we want to run the test also as part of the build process (only if compiling with python support of course), in this way if the python API breaks we will know right away? 

regarding code coverage, I guess that we will want something that will support also c++ (I am not familiar with the tools you suggested). but anyway, sounds like a different task, so we can finish this one and then do that

kjulian3:
I haven't worked with travis cache before. Is there a way to cache only for windows? Also I found online that we might be able to cache the python packages installed with pip using "cache: pip", but they say that language: python must be set. I'm not sure if it works with language: cpp.

I've add the tests to cmake along with a flag to turn on or off the tests. I updated the test file so the tests can be run from any directory.

CMakeLists.txt
```diff
@@ -268,6 +269,22 @@ if (NOT ${TESTS_TO_RUN} STREQUAL "")
...
+        )
+    endif()
+
+    add_custom_command(
+        TARGET build-tests
+        POST_BUILD
+        COMMAND ${PYTHON_EXECUTABLE} -m pytest ${PYTHON_API_DIR}/test
+    )
+endif()    
+
```
yuvaljacoby, line 287: Did you test this on windows and unix?  shouldn't the second command be in else statement?

kjulian3: The second command runs the python api tests, which should be done for both windows and unix. Windows puts the MarabouCore shared library in a Release subdirectory, which needs to be copied to the maraboupy directory to be used. Once it's copied, the tests run the same way for all build types. I tested on my laptop, and travis was able to build correctly. The travis outputs show that the tests were run and passed after the build.

yuvaljacoby:
Sorry for the late response.

I don't think we can cache only for windows (you can try to put an if statement there) 
But maybe we can cache an empty directory in unix? 

I will be able to take a deeper look only in about a week in a half

CMakeLists.txt
yuvaljacoby: oops

kjulian3:
I'll merge for now so I can start adding more test cases. We can look into adding to the travis cache in a future pr.

PR #258, opened 2020-05-03
Small fixes for using NLR with python API

kjulian3:
I couldn't get NLR to run with in the python API, so I made a few changes to get it working. I added use_nlr to the Marabou.read_nnet function so it can be used easily. I added the ActivationFunction enum to MarabouCore, because I was getting complaints that an int wasn't an acceptable last input to nlr.setNeuronActivationFunction(). Lastly, the layer index needs a +1 so we start with the hidden layer and not the input layer.

I've attached a script that checks a small region of one of the ACAS Xu prototype networks, where the minimal cost action should always be the third with zero-based indexing. The script verifies this with nlr off and nlr on, and with nlr on the runtime is reduced to less than half the time when nlr is off. The script needs to run from the Marabou directory, where resources is a sub-directory.

[TestNetwork.zip](https://github.com/NeuralNetworkVerification/Marabou/files/4569720/TestNetwork.zip)


guykatzz:
Good catch with the indexing, thanks. And also thanks for adding the ENUM - I intended to go back and fix it, and then forgot about it :).

PR #263, opened 2020-05-17
travis speedup, cache python3 on windows, use cache for pip packages

yuvaljacoby:


kjulian3:
Looks good to me, except python and pip install commands seem to have run without using cached directories. Is this because the directories weren't cached yet? Will the cached directories be available when travis runs in the future? Is there a way to check what happens when the directories have been cached and make sure that using the cached directories works correctly?

yuvaljacoby:
pip - the cache is only for downloading the packages, still need to install them (which takes most of the time)
python - yes, it downloaded again only because no cache directory was available, to make sure of that I restarted the build (on travis)

kjulian3:
Looks like the caching works, great!

PR #265, opened 2020-05-19
Setup codecov for maraboupy

kjulian3:
This sets up codecov for the maraboupy python files. The new file pytest.ini configures pytest calls to include code coverage of the maraboupy folder. The new file .coveragerc controls the options for the python code coverage, such as which files should be omitted. Other options can be added in the future if needed.


kjulian3:
The codecov report can be found here: https://codecov.io/gh/NeuralNetworkVerification/Marabou

Seems like the default behavior is to comment on each pull request with how the code coverage changes with the pull request. Right now the master doesn't have a codecov report, so the comparison isn't helpful. Do we want to keep these reports automatically commenting on each pull request? That seems annoying an unnecessary to me.

We'll have to make more changes to setup C++ code coverage, but I think we can save that for the next pull request, unless it's easy to add.

yuvaljacoby:
* I don't think we need to push a report on every pull request
* I agree that we can dell with c++ later


I think we should add a badge to the README file (similar to [CVC4](https://github.com/CVC4/CVC4) ) which I hope will encourage us to add more tests :) 

I will take a look on maraboupy/tests.py later and add it to the "suite"
@anwu1219 do we have any DNC tests ?

kjulian3:
I switched to codecov.yml instead of .coveragerc, which gives us the option to disable pr comments as well as access to more options that may be helpful for integrating c++ code coverage in the future.

Issue #267, opened 2020-05-20
CMake runs python tests but does not check for or install required packages


kjulian3:
Python tests were added to CMake in pr #254, but we don't check that the required packages are installed before running the test. Therefore, if a user does not have pytest installed, for example, the build will fail.

yuvaljacoby:
I think this is tricky,
We need to distinguish between two types of users:
* developers  - which I think is better to let them install packages the way they like
* users - which I think should be able to _pip install_, I plan to make that available in the next couple of weeks

So I would not address this issue, @ahmed-irfan @guykatzz @AleksandarZeljic  any thoughts?

kjulian3:
In the mean time, should we set RUN_PYTHON_TEST to OFF as the default in CMake?

We could test maraboupy after building without pytest, and use only nnet format examples, which would let us test maraboupy without needing to install pytest, pytest-cov, codecov, tensorflow, onnx, or onnxruntime. We would still need to install numpy or check that it is installed, but most python users have numpy installed already, and essentially any version of numpy would suffice.

We could also create and run tests using MarabouCore directly, which would avoid the use of numpy, like this example: https://github.com/NeuralNetworkVerification/Marabou/blob/master/maraboupy/examples/MarabouCoreExample.py

yuvaljacoby:
Maybe keep the flag on but change the README to disable the tests (cmake .. -DRUN_PYTHON_TEST=off) ? 
We still run  tests on travis so nothing "bad" can happen .... 

I am trying to think of something that will keep it simple but we will want to fix soon.

yuvaljacoby:
closing this issue,  we have #192 to enable pip install

PR #268, opened 2020-05-20
Pytest fixes

kjulian3:
This pr fixes two minor issues. First, pytest.ini was used so that code coverage was checked every time pytest runs. However, when running pytest during the build process, code coverage is not needed. This is fixed by deleting the pytest.ini file and explicitly calling for code coverage during the travis build with "--cov=maraboupy".

Second, this pr address issue #267 by modifying cmakelists to install python packages required for testing via pip. Packages are installed via pip if the options for BUILD_PYTHON and RUN_PYTHON_TESTS are set to ON (which is the current default). The maraboupy/test_requirements.txt file is used to specify the versions of these packages that should be installed.

The downside is that this could cause the versions of a user's python packages to upgrade or downgrade, which they might not want if they need a particular version for other packages. I'm not sure how to avoid this if we want to run the python tests automatically after building. On the other hand, these package versions are known to work correctly with maraboupy, so this could help users get the package versions needed to run maraboupy.


yuvaljacoby:
The first fix is great.
I wouldn't do the second (comments in issue #267 ) 
Let's close this PR with only the first fix and discuses in the issue if we want to solve it or not

kjulian3:
I removed the pip install from cmake. I also turned off the python tests by default so that users can build Marabou without having required python packages installed, as was done before the tests were added.

Issue #269, opened 2020-05-20
MaraboupyTensorflow parser errors for Tensorflow 2.0


kjulian3:
The tensorflow parser was written before Tensorflow version 2.0 came out in the fall of 2019, and version 2.0 introduced many changes and deprecations. Luckily, Tensorflow created a script to automatically update legacy code, so updating files to be compatible with version 2.0 should be easy:
https://www.tensorflow.org/guide/upgrade

This issue can be reproduced by installing tensorflow version 2.0 or greater and running this python code from the root Marabou directory:
```
from maraboupy import Marabou
network = Marabou.read_tf("maraboupy/examples/networks/graph_test_medium.pb")
```

PR #270, opened 2020-05-20
Support for tensorflow 2.0 and initial tests for tensorflow parser

kjulian3:
Addresses #269 by making the minimal changes to support tensorflow 2.0. Older versions of tensorflow will still work as well.

This pr also adds tests for the tensorflow parser to validate that the changes made do not cause errors. Travis currently runs tensorflow v1.15.2, but the tests can also be run with versions 2.0 or greater. The tensorflow tests follows the same style as the ONNX tests.

These changes also eliminate some warnings produced by tensorflow, so two of the warning filters can be removed, affecting both ONNX and tensorflow tests.


kjulian3:
The code coverage for the code changed by this "patch" is low because there are no tests for DNC mode yet. @anwu1219  do you have any tests to add for DNC mode?

Also, I'm having trouble writing test cases in the SavedModel format. Tensorflow 2.0 changes this format and methods used to save/load the format. I don't know if anyone uses the SavedModel format, and it may be best to just delete the format and only support the frozen network format.

anwu1219:
Hi Kyle, could you try to merge from the master? There will be a merge conflict since NumTasksDecider is removed.

maraboupy/Marabou.py
```diff
@@ -42,7 +42,7 @@ def read_nnet(filename, use_nlr=False):
     return MarabouNetworkNNet(filename, use_nlr=use_nlr)
 
 
-def read_tf(filename, inputNames=None, outputName=None, savedModel=False, savedModelTags=[]):
+def read_tf(filename, inputNames=None, outputName=None, savedModel_v1=False, savedModel_v2=False, savedModelTags=[]):
```
anwu1219, line 45: is it possible to merge saveModel_v1, saveModel_v2 as a single argument called for instance "modelFormat". which is a string that can be one of  "frozen", "savedModelv1", "savedModelv2"?

kjulian3: Yes I think that's a better way to do it. I'll work on those changes. I'm also working on creating tests for savedModelv1, so there's still a bit of work to do before this is ready to be merged.

kjulian3:
I added tests for SavedModel format created from tensorflow versions 1.15.2 and 2.2.0.  All tests pass when using tensorflow version 2, and I also ran the tests locally for tensorflow versions 1.13.1 and 1.15.2, and all tests except for the savedModel_v2 pass. So the tensorflow parser can be used with tensorflow versions 1 or 2, version 2 users can read models created from either tensorflow v1 or v2, and version 1 users can read models created from tensorflow v1.

PR #271, opened 2020-05-20
NLR for networks with a general, feed-forward topology

guykatzz:
The motivation for this PR is to have the network level reasoner support networks where neurons from layer i may be connected to any layer j > i, not necessarily j=i+1. Further, we want to support any layer ordering - e.g., cases where we have multiple ReLU layers with no weighted sums.

The main changes are:
- NLR revised significantly, and moved to its own folder and name space
- Support symbolic bound tightening and interval arithmetic bound tightening for these arbitrary feed forward networks.
- Support the auto construction of an NLR for these networks (done in the preprocessor)
- The layer indexing now assigns a unique index to every weighted sum layer and every activation layer, whereas previously weighted sum + activation were grouped into one index.

Other, smaller changes:
- Don't eliminate input/output variables in the preprocessing pass
- Have the NLR store the PLCs in a topological order, to be used by splitting heuristics

I compared performance on the master branch. Runtimes and state numbers are unchanged, which indicates that the SBT functionality has not been damaged.


yuvaljacoby:
I had hard time following back and forth between the NetworkLevelReasoner  and Layer objects.
Not sure why we need the Layer object in a different context, and if so I guess we should move it elsewhere (common?)  

besides that there are some code repetition and small stuff that can be fixed along the way (for example - in Layer.cpp lines: 534-540 is similar to 698-704 initialize bounds using _eliminatedNeurons)

guykatzz:
Hi Yuval,
The NLR class itself serves as a container for the Layer objects, where the actual heavy lifting is done. This is just a modular design that reflects the way, e.g., symbolic bound tightening is performed: each time a single layer is computing tighter bounds. Perhaps the name Layer isn't great, although the name space NLR::Layer helps a bit. I don't think that it belongs in common, because it performed network level operations.

I agree that there is some small code duplication where you pointed, but felt that this afforded better readability.

PR #276, opened 2020-05-27
move print statement in NLR construction to logger

anwu1219:
- Create a log function in the Preprocessor class.
- Move the print statements to logger in the Preprocessor::preprocess method.


grese:
👍

PR #284, opened 2020-06-11
update the statistics object even if verbosity is 0

anwu1219:
When verbosity is set to 0, _statistics in Engine.h is not updated. However, the precision restoration depends on _statistics. Therefore, we update the _statistics even if we do not print it out.


kjulian3:
This fixes the issue I was seeing where changing the verbosity results in different SAT points found.

PR #287, opened 2020-06-14
Maraboupy core tests

kjulian3:
The file maraboupy/tests.py was moved and refurbished to create the test_core.py, which tests that Marabou can be called from python by interacting directly with MarabouCore. The test can be run with only pytest installed. Moving MarabouNetworkNNet import into a try block allows Maraboupy to work even when numpy is not installed.

This is a starting point for testing MarabouCore, but there are additional functions not yet tested, such as
* DNC mode
* createInputQuery()
* Additional PL constraints, such as absolute value and disjunction
* Statistics methods beyond hasTimedOut()

In addition, MarabouCore.cpp is not yet part of code coverage


maraboupy/test/test_core.py
```diff
@@ -0,0 +1,81 @@
...
+    ipq = MarabouCore.InputQuery()
+    ipq.setNumberOfVariables(3)
+
+    # x
+    ipq.setLowerBound(0, -1)
+    ipq.setUpperBound(0, 1)
+
+    # relu(x)
+    ipq.setLowerBound(1, 1)
+    ipq.setUpperBound(1, 2)
```
yuvaljacoby, line 60: should be 0,1 ? otherwise you set x and relu(x) to be always 1

kjulian3: That would make more sense, and I changed it to 0, LARGE because the upper bound comes from the upper bound on the first variable, which Marabou should propagate. I've reformatted the tests a little bit to also check that Marabou provides a default bounds when none are provided, and that bound propagation tightens these bounds as well.

kjulian3:
Yeah, codecov is mad that I didn't write a test for the small change I made to Marabou.py. To test that part of the code, we'd have to run tests without numpy installed. I've tested it on my machine, and I don't think it's worth trying to configure tests to run without numpy to show it works.

PR #289, opened 2020-06-15
Cleaning examples folder

kjulian3:
This pr updates the examples folder and removes outdated files.

In addition, this pr makes the ONNX and tensorflow parsers consistent in that the output variables preserves the batch_size dimension. Often, ONNX/tensorflow networks have an extra first dimension that represents the batch size, or number of input points to be evaluated. The tensorflow parser removes that extra dimension, but that causes the output shape to differ from ONNX. I think it's safest to just keep the extra dimension and let users handle it as needed.


kjulian3:
Also, I'm confused why debug.ipynb is in the examples folder. @GuyKatzHuji, do you know why that file is present?

PR #290, opened 2020-06-16
keep log lines only in debug mode

yuvaljacoby:
Change all log functions into define's 
Remove log define's in Release mode (deletes the whole line)

### DIFF
--- a/src/common/Debug.h
+++ b/src/common/Debug.h
@@ -27,19 +27,19 @@
 #endif
 
 #ifndef NDEBUG
-#  define LOG(x, f, y...)                     \
+#  define LOG(x, f, y...)                       \
     {                                           \
-        if ( ( x ) )                           \
+        if ( ( x ) )                            \
         {                                       \
-            printf( f, y );            \
-       }                                       \
+            printf( f, y );                     \
+       }                                        \
     }
 #else
 #  define LOG(x, y, ...)  {}
 #endif
 
 #ifndef NDEBUG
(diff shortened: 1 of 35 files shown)


src/common/Debug.h
```diff
@@ -26,7 +27,19 @@
 #endif
 
 #ifndef NDEBUG
-#  define ASSERTM(x, y, ...)                     \
+#  define LOG(x, f, y, ...)                     \
+    {                                           \
+        if ( ( x ) )                            \
+        {                                       \
+            printf( f, y );                     \
+       }                                        \
```
guykatzz, line 35: Indentation

src/common/Debug.h
```diff
@@ -26,7 +27,19 @@
...
+    }
+#else
+#  define LOG(x, y, ...)  {}
```
guykatzz, line 38: Why (x, y, ...) and not (x, f, y, ...)? For symmetry

src/engine/ProjectedSteepestEdge.cpp
```diff
@@ -125,7 +125,7 @@ bool ProjectedSteepestEdgeRule::select( ITableau &tableau,
 
     if ( remainingCandidates.empty() )
     {
-        log( "No candidates, select returning false" );
+        PSEU_LOG( "No candidates, select returning false" );
```
guykatzz, line 128: Why PSEU and not PSE?

Issue #190, opened 2019-10-16
Loading Pytorch Models


darioguidotti:
Hello,

I am trying to find a way to verify a set of models created and 
learned using pytorch but I am finding some difficulties.

Converting the models to the ONNX format and then to tensorflow protobuffer
(using onnx-tf) seems to modify the model with unnecessary operations which
are currently not supported by Marabou (e.g., GatherV2).

My models are convolutional and fully connected networks which only use Conv2d, Linear and MaxPool2d layers and the torch reshape function in forward of the model.

Do you have any advice on how to solve this problem?

Thank you!

kjulian3:
Support has been added for the tensorflow network layers in this network. An ONNX parser has been created, which can handle networks translated to ONNX from pytorch.

Issue #291, opened 2020-06-17
MaxConstraint assertion violation


kjulian3:
I've noticed a strange issue that sometimes happens when using Maraboupy to evaluate a random point of a network that uses with max-pool. Sometimes, Marabou will reach an assertion violation on line 275 of MaxConstraint.cpp. The error seems to occur during pre-processing, but it rarely errors. In the script below, I'm generating the same input query and solving it 500 times, and usually at some point before reaching 500 an assertion violation occurs.

The error occurs in both Release and Debug modes. I've done a little debugging, and found that the resetMaxConstraint() function is getting called when _assignments is empty while _elements has two or three entries. When Marabou does not error, resetMaxConstraint is never called; only when maxErased is true and resetMaxConstraint is called does Marabou violate the assertion. I'm not sure how to debug the issue further and would appreciate some help.

This issue has sometimes caused Travis's python tests to fail, though rerunning the tests usually succeeds since the failure happens rarely.

Example python code to reproduce the issue:
```
from maraboupy import Marabou
import numpy as np
import os

# Global settings
OPT = Marabou.createOptions(verbosity = 2)  # Turn off printing
FG_FOLDER = "resources/tf/frozen_graph/"    # Folder for test networks written in frozen graph format
RANDOM_SEED = 1234                          # Seed random numbers for repeatability

"""
Test a convolutional network using max pool
"""
filename = os.path.join(os.path.dirname(__file__), FG_FOLDER, "conv_mp2.pb")
network = Marabou.read_tf(filename)
np.random.seed(RANDOM_SEED)
testInput = [np.random.random(inVars.shape) for inVars in network.inputVars]
for i in range(500):
    network.evaluateWithMarabou(testInput, options = OPT, filename = "")
    print(i)
```

kjulian3:
I've also been able to reproduce this issue with an onnx network by setting 
```
filename = "resources/onnx/conv_mp1.onnx"
network = Marabou.read_onnx(filename)
```
in place of the tensorflow filename and read_tf functions. The onnx network is smaller, with fewer variables and equations, so it may be a better test network.

I've tried saving the input query using MarabouCore.saveQuery(), but the resulting file from a query that caused an error is identical to one that was produced by a query that didn't cause an error. I have also been unable to reproduce the error when using MarabouCore.loadQuery() to load the query file and solve with Marabou, even when it was written by a query that caused a failure.

kjulian3:
Solved with #297

PR #292, opened 2020-06-17
Multiply/divide in tensorflow

kjulian3:
Adding support for element-wise multiply/divide allows us to solve #190. We check that the network isn't multiplying/dividing two variable arrays together, and that the network isn't dividing a constant by a variable. Otherwise, implementation is straight-forward. A test network was added as well.


maraboupy/test/test_tf.py
```diff
@@ -44,7 +53,7 @@ def test_conv_mp1():
     """
     filename = os.path.join(os.path.dirname(__file__), FG_FOLDER, "conv_mp1.pb")
     network = Marabou.read_tf(filename)
-    evaluateNetwork(network, numPoints = 5) 
+    evaluateNetwork(network, numPoints = 3)
```
anwu1219, line 56: Why do we reduce the number of sampled points?

maraboupy/test/test_tf.py
```diff
@@ -80,7 +89,7 @@ def test_conv_NCHW():
     network_nchw = Marabou.read_tf(filename)
 
     # Evaluate test points using both Marabou and Tensorflow, and assert that the max error is less than TOL
-    testInputs = [[np.random.random(inVars.shape) for inVars in network_nhwc.inputVars] for _ in range(NUM_RAND)]
+    testInputs = [[np.random.random(inVars.shape) for inVars in network_nhwc.inputVars] for _ in range(5)]
```
anwu1219, line 92: Maybe reduce the value of NUM_RAND instead?

kjulian3:
No, the failure comes from something within Marabou. If I were to re-run Travis, the test would likely pass. I documented this issue in #291. If you run the script provided in that issue, are you able to reproduce this strange failure?

maraboupy/test/test_tf.py
kjulian3: I reduced the number of points for the convolutional networks that are much larger than other test networks so that we don't add too much time to Travis. I can just reduce NUM_RAND to 5 for all tests, and the tests run in about 18 seconds.

kjulian3:
Looks like Travis fails because of lcov: https://travis-ci.com/github/NeuralNetworkVerification/Marabou/jobs/351539987

Even though the issue with lcov was fixed in #299, those changes to master are not getting pulled in when Travis runs, even if the travis build is restarted. We can tell this is the case since the travis config does not match that current master .travis.yml. Is there a way to pull in changes to master and re-run travis?

maraboupy/test/test_tf.py
kjulian3: I lowered NUM_RAND  to 5

kjulian3:
Fixed by rebasing this pr branch on upstream/master

PR #293, opened 2020-06-17
Cpp codecov

yuvaljacoby:


kjulian3:
This looks good to me. Still travis sometimes fails randomly for max-pool networks used as maraboupy tests. I tried to capture that issue in #291, but it's difficult to reproduce.

PR #296, opened 2020-06-18
move input query construction from preprocessor to inputquery

anwu1219:
This allows us to call constructNLR() after loading the .nnet file, so that we can load bounds of internal variables in the property file.


src/configuration/GlobalConfiguration.cpp
```diff
@@ -89,7 +89,7 @@ const bool GlobalConfiguration::TABLEAU_LOGGING = false;
 const bool GlobalConfiguration::SMT_CORE_LOGGING = false;
 const bool GlobalConfiguration::DANTZIGS_RULE_LOGGING = false;
 const bool GlobalConfiguration::BASIS_FACTORIZATION_LOGGING = false;
-const bool GlobalConfiguration::PREPROCESSOR_LOGGING = false;
+const bool GlobalConfiguration::INPUT_QUERY_LOGGING = true;
```
guykatzz, line 92: Set to false by default?

anwu1219: Done!

guykatzz:
LGTM

src/input_parsers/PropertyParser.cpp
```diff
@@ -138,7 +138,7 @@ void PropertyParser::processSingleLine( const String &line, InputQuery &inputQue
 
             auto subToken = subTokens.begin();
             ++subToken;
-            unsigned layerIndex = atoi( subToken->ascii() );
+            unsigned layerIndex = 2 * atoi( subToken->ascii() ) - 1;
```
guykatzz, line 141: Hmm...
So, it seems like this change is tightly coupled to the ACAS Xu architecture, of input, <fc,relu>*, output layers. Our NLR now supports general feed-forward topologies. I understand the motivation was to maintain the old ws indexing, but I'm not sure it's good to hard-code this.

PR #305, opened 2020-06-24
Copyright

guykatzz:


mpreiner:
@guykatzz Please do not remove the copyright lines if you adapt scripts from the CVC4 project. Leave the CVC4 project copyright in the header and add the Marabou project copyright if the Marabou project modified the files.

PR #310, opened 2020-06-28
Add features to MarabouNetworkNNet.

alexopenu:
- Add write to file to MarabouNetworkNNet.

- Add evaluate without Marabou.

- Create random input for MarabouNetworkNNet.

- Add tests.

- Add normalization to evaluate without Marabou.


maraboupy/MarabouNetworkNNet.py
```diff
@@ -152,108 +257,303 @@ def read_nnet(self, file_name):
...
-        input_variables = []
+
+        """
+        Compute the variable number ranges for each type (input, output, b, f)
+
+        Args:
+            None
+        """
+
+        # input_variables = []
```
guykatzz, line 376: Please remove these commented-out lines

maraboupy/MarabouNetworkNNet.py
```diff
@@ -263,31 +563,346 @@ def buildEquations(self):
...
+        """
+         Evaluate network using multiple sets of inputs
+
+         Args:
+             inputs (numpy array of lists of floats): Array of network inputs to be evaluated.
+
+         Returns:
+             (numpy array of numpy arrays floats): Network outputs for each set of inputs
+
+        HAS NOT BEEN TESTED
```
guykatzz, line 659: Please either test or remove this function

guykatzz: Also, why not invoke the previous functions multiple times? This seems like code duplication

maraboupy/MarabouNetworkNNet.py
```diff
@@ -263,31 +563,346 @@ def buildEquations(self):
...
+
+        # Undo output normalization
+        if (normalize_outputs):
+            for i in range(outputSize):
+                for j in range(numInputs):
+                    outputs[i, j] = outputs[i, j] * self.outputRange + self.outputMean
+
+        return outputs.T
+
+    def evaluateNetworkToLayer(self, inputs, last_layer=-1, normalize_inputs=False, normalize_outputs=False,
```
guykatzz, line 706: Also seems to duplicate the previous functions - can we unify them?

maraboupy/MarabouNetworkNNet.py
```diff
@@ -263,31 +563,346 @@ def buildEquations(self):
...
+        numLayers = self.numLayers
+        outputSize = self.outputSize
+        biases = self.biases
+        weights = self.weights
+        mins = self.inputMinimums
+        maxes = self.inputMaximums
+        means = self.inputMeans
+        ranges = self.inputRanges
+
+        # Prepare the inputs to the neural network
```
guykatzz, line 796: Remove commented-out code

maraboupy/MarabouNetworkNNet.py
```diff
@@ -263,31 +563,346 @@ def buildEquations(self):
...
+        """
+         Evaluate network directly, up to a certain layer, using multiple sets of inputs
+
+         Args:
+             inputs (list of floats): Array of network inputs to be evaluated.
+
+         Returns:
+             (numpy array of floats): Network outputs for each set of inputs
+
+        HAS NOT BEEN TESTED
```
guykatzz, line 839: Same

maraboupy/MarabouNetwork.py
```diff
@@ -287,3 +289,26 @@ def findError(self, inputValues, options=None, filename="evaluateWithMarabou.log
         outNotMar = self.evaluate(inputValues, useMarabou=False, options=options, filename=filename)
         err = np.abs(outMar - outNotMar)
         return err
+
+    def evaluateWithoutMarabou(self, inputValues):
```
kjulian3, line 293: Defining this function in MarabouNetwork.py is not necessary, as it can be defined in the child class, MarabouNetworkNNet.py. See how evaluateWithoutMarabou() is already defined in the tensorflow and onnx parser files.

maraboupy/MarabouNetworkNNet.py
```diff
@@ -49,13 +54,111 @@ def __init__ (self, filename, normalize=False):
...
+            lowerBounds = dict()
+            upperBounds = dict()
+            inputVars = []
+            outputVars = np.array([])
         """
         super().__init__()
 
         self.normalize = normalize
+
+        if not filename:
```
kjulian3, line 81: This could be reduced to self.clearNetwork()

maraboupy/MarabouNetworkNNet.py
```diff
@@ -49,13 +54,111 @@ def __init__ (self, filename, normalize=False):
...
         self.read_nnet(filename)
 
+        self.computeNetworkAttributes()
+
+    def clearNetwork(self):
+        # call clear() from parent
+        self.clear()
+
+        self.numLayers = 0
+        self.layerSizes = 0
```
kjulian3, line 109: self.layerSizes should default to an empty list

maraboupy/MarabouNetworkNNet.py
```diff
@@ -49,13 +54,111 @@ def __init__ (self, filename, normalize=False):
...
+        self.outputSize = 0
+        self.maxLayersize = 0
+        self.inputMinimums = []
+        self.inputMaximums = []
+        self.inputMeans = []
+        self.inputRanges = []
+        self.weights = []
+        self.biases = []
+
+    def resetNetworkFromParameters(self, inputMinimums: list, inputMaximums: list, inputMeans: list, inputRanges: list,
```
kjulian3, line 120: What's the purpose of resetNetworkFromParameters? This changes some attributes, but not the equations, variables, or constraints.

maraboupy/MarabouNetworkNNet.py
```diff
@@ -263,31 +563,346 @@ def buildEquations(self):
...
+        return self.layerSizes[0] + 2 * sum(self.layerSizes[1:-1]) + 1 * self.layerSizes[-1]
 
     def getInputMinimum(self, inputVar):
         return self.inputMinimums[inputVar]
 
     def getInputMaximum(self, inputVar):
         return self.inputMaximums[inputVar]
+
+
+    def evaluateNetwork(self, inputs, normalize_inputs=True, normalize_outputs=True, activate_output_layer=False):
```
kjulian3, line 596: As mentioned in a previous comment, this function should be named "evaluateWithoutMarabou". Variables normalize_inputs and normalize_outputs should be changed to self.normalize, and activate_output_layer is not necessary as the NNet format assumes activation only on the hidden layers, not output layer. This function should only take "inputs" as the argument (besides "self").

maraboupy/test/test_nnet_extensions.py
```diff
@@ -0,0 +1,149 @@
...
+property_filename = "./resources/properties/acas_property_4.txt"
+network_filename = "./resources/nnet/acasxu/ACASXU_experimental_v2a_1_9.nnet"
+
+layer = 2
+
+
+def test_write_to_file():
+    output_filename = "./maraboupy/test/ACASXU_experimental_v2a_1_9_output.nnet"
+
+    nnet_object = MarabouNetworkNNet(filename=network_filename)
```
kjulian3, line 16: It's better practice to use Marabou.read_nnet( filename ) instead of calling MarabouNetworkNNet directly. Users only need to import maraboupy.Marabou to do anything they want with Maraboupy, whether it's reading an NNet file or tensorflow file, and these tests should be a good model for how Maraboupy should be used.

maraboupy/test/test_nnet_extensions.py
```diff
@@ -0,0 +1,149 @@
...
+from subprocess import call
+
+property_filename = "./resources/properties/acas_property_4.txt"
+network_filename = "./resources/nnet/acasxu/ACASXU_experimental_v2a_1_9.nnet"
+
+layer = 2
+
+
+def test_write_to_file():
+    output_filename = "./maraboupy/test/ACASXU_experimental_v2a_1_9_output.nnet"
```
kjulian3, line 14: A temporary file can be used. See the first test in test_query.py for an example. A "tmpdir" argument is added to the test method, which can be used to make a temporary file for writing the nnet file.

maraboupy/test/test_nnet_extensions.py
```diff
@@ -0,0 +1,149 @@
...
+
+    nnet_object = MarabouNetworkNNet(filename=network_filename)
+    nnet_object.writeNNet(output_filename)
+
+    call(['diff', output_filename,network_filename])
```
kjulian3, line 19: Doesn't this need some sort of assert?

maraboupy/test/test_nnet_extensions.py
```diff
@@ -0,0 +1,149 @@
...
+        without_marabou_output = nnet_object_a.evaluate(np.array([inputs]),useMarabou=False).flatten()
+        assert (output2 == without_marabou_output).all()
+
+        with_marabou_output = nnet_object_a.evaluate(np.array([inputs]), useMarabou=True).flatten()
+        with_marabou_output_rounded = np.array([float(round(y, 8)) for y in with_marabou_output])
+        output2_rounded = np.array([float(round(y, 8)) for y in output2])
+        assert (output2_rounded == with_marabou_output_rounded).all()
+
+
+if __name__ == "__main__":
```
kjulian3, line 145: For consistency, it would be good to follow the same type of format as the other test files and use pytest, in which case this main function is not needed.

maraboupy/test/test_nnet_extensions.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # Compare evaluation with and without Marabou
+        without_marabou_output = nnet_object.evaluate(np.array([inputs]),useMarabou=False).flatten()
+        without_marabou_output_rounded = np.array([float(round(y,8)) for y in without_marabou_output])
+
+        with_marabou_output = nnet_object.evaluate(np.array([inputs]),useMarabou=True).flatten()
+        with_marabou_output_rounded = np.array([float(round(y,8)) for y in with_marabou_output])
+
+        # Assert that all of the above agree, at least up to 10^-8
+        assert (output1 == without_marabou_output).all()
```
kjulian3, line 44: Rather than rounding and asserting equality, the absolution value of the difference can be asserted to be less than some tolerance value (such as 1e-8). Making tolerance a global variable allows the value to be easily changed for all asserts. See the other test files as an example.

maraboupy/MarabouNetworkNNet.py
kjulian3: The evaluate...Multiple functions are not necessary. These functions seem to be mostly copied from the NNet repository, and code for evaluating multiple points at once was implemented in the NNet repository to improve computation time when evaluating a large number of points. For Marabou, I don't think there is a need for this multiple-evaluation function since we just want to check a few points to show that Marabou equations produce the same results as an NNet evaluation.

maraboupy/MarabouNetworkNNet.py
kjulian3: Inputs to this function such as first_layer and last_layer that default to -1, for example, would allow you to combine this function and the to/from layer evaluation functions into one.

maraboupy/MarabouNetworkNNet.py
```diff
@@ -263,31 +563,346 @@ def buildEquations(self):
...
+            else:
+                output_mean = means[layer - 1]
+                output_range = ranges[layer - 1]
+
+            for i in range(outputSize):
+                outputs[i] = outputs[i] * output_range + output_mean
+
+        return outputs
+
+    def evaluateNetworkFromLayer(self, inputs, first_layer=0, normalize_inputs=False, normalize_outputs=False,
```
kjulian3, line 774: What is the use case for this function? Marabou equations are currently not constructed to start or end in an intermediate layer, though that functionality could be added by adding arguments "first_layer" and "last_layer" to the __init__ function. Without that functionality implemented, I don't see how this function is useful for Marabou.

maraboupy/MarabouNetworkNNet.py
alexopenu: I have originally considered this, but I generally prefer to include initiation of all the attributes in __init__

However, if it seems preferable to you, I don't feel strongly in this particular case, I think it will be clear enough either way.

maraboupy/test/test_nnet_extensions.py
alexopenu: I have removed this test. 

writeNNet is tested in a different test via evaluation of the two networks; here I just wanted to see what the diff looks like.

maraboupy/test/test_nnet_extensions.py
alexopenu: Done. 

BTW, is there a reason why some of the imports were relative? (e.g., from .Marabou import.. )? I have changed all of them to maraboupy.Marabou etc, otherwise things don't work on my system (but I can try to fix this issue, if the relative import is important).

maraboupy/test/test_nnet_extensions.py
alexopenu: Done.

maraboupy/MarabouNetwork.py
alexopenu: Done.

maraboupy/MarabouNetworkNNet.py
alexopenu: My bad. Thanks!

maraboupy/MarabouNetworkNNet.py
alexopenu: I think it does: it calls self.clear() [from MarabouNetwork] to clear these attributes, and then self.computeNetworkAttributes() to recompute the new ones.

maraboupy/MarabouNetworkNNet.py
alexopenu: I have now combined all the evaluations in one method evaluateNNet, and moved evaluateWithoutMarabou (which calls evaluateNNet) from MarabouNetwork to MarabouNetworkNNet, as you have suggested. 

The reason I want to have the additional features has to do with a project that I am working on (invariant search), which involves looking at the behavior of different hidden layers, and evaluating the network to a from a certain layer, as well as splitting the network into two (I have a function that does this for an nnet, which I did not include in the PR). In particular, sometimes I want to activate the output layer (of a "sliced" network) in order to get correct results.

maraboupy/MarabouNetworkNNet.py
alexopenu: Actually, as a part of computing statistics for hidden neurons, as well as for checking a candidate for an invariant (without invoking Marabou), I do evaluate the networks on tens of thousands of inputs at the time. So I thought that I eventually I would use evaluate...Multiple for this, precisely for improving efficiency. At the moment I have removed them from this PR. When and if I start using them, I can add them back (together with appropriate tests).

maraboupy/MarabouNetworkNNet.py
alexopenu: You are right. Once upon a time, it made sense to have one method with straightforward functionality that I could compare all the other ones to in order to make sure that I am not doing anything stupid while adding different functionalities, but now this is all unnecessary. I have now combined all the evaluate methods into one (with optional parameters), and removed (for now) evaluate multiple. So there should be any code duplication.

maraboupy/MarabouNetworkNNet.py
alexopenu: I use this just for evaluating the network between given layers, not evaluating Marabou equations. As I have mentioned in a previous comment, this is very helpful for invariant e.g. search. 

Thanks for all the comments, very helpful. I think I've addressed all your concerns, would you mind having another look?

maraboupy/MarabouNetworkNNet.py
kjulian3: I think it's cleaner to define the attributes in one place. It will also make sure we don't define the attributes in different ways in each place, and will make it easier to update in the future since we only define the default attributes in one place.

maraboupy/MarabouNetworkNNet.py
kjulian3: Ok I see that now. So if I understand correctly, this function allows users to supply their own parameters instead of reading an NNet file, which is helpful if they have a network in a different file format, for example? If so, I think we should make the input/output min/max/mean/range arguments optional. Using mean of 0 and range of 1 effectively skips normalization. Input min/max could default to float min/max, or just set to [] and edit other functions that use inputMinimimums/inputMaximums adjusted to handle an empty vector.

maraboupy/MarabouNetworkNNet.py
kjulian3: I see, thanks for the explanation! A function to split the network into two might be a useful feature for a subsequent pr.

maraboupy/test/test_nnet_extensions.py
kjulian3: Sometimes the relative imports are helpful if the Marabou repo has not been added to PYTHONPATH, and the user is not in the Marabou repo folder. What errors were you seeing with relative imports? You're right that we are not consistent about doing it one way or another, which is something we should address in the future.

maraboupy/MarabouNetworkNNet.py
kjulian3: No problem, thanks for the work done to add new features! I'll take another look through now.

maraboupy/test/test_nnet_extensions.py
alexopenu: They just were not working for me for some reason (python did not find the modules). I think that it would be good to agree on a uniform policy, and if we decide to stick with relative imports, I will see how to fix the issue on my Mac/PyCharm combination.

Speaking of uniformity: there are several (but very few) methods that are named using a different (perhaps a more "pythonian") naming convention, for historical reason, I assume: read_nnet, for instance. Should we consider changing this, maybe? Or will this cause too much trouble in too many scripts of too many people in too many different repositories at this point?

maraboupy/MarabouNetworkNNet.py
alexopenu: Guy thought that it did not belong in the master at this point, this is why I excluded it from this PR. :) But if you think that it might be useful, perhaps I will still include it in the next one.

maraboupy/MarabouNetworkNNet.py
alexopenu: Exactly, or if the network is created by splitting or slicing an already existing network represented by a MarabouNetworkNNet object. 

Making some of the arguments optional is an excellent idea! As a matter of fact, in the split-network function, some of them are currently being set to these defaults anyway (e.g. mean=0, range=1, and float min/max, precisely as you say). So it would be good to make this more systematic. I will make these changes.

maraboupy/MarabouNetworkNNet.py
alexopenu: Yes, I think that you are right. I will change this.

maraboupy/MarabouNetworkNNet.py
alexopenu: Changes made. Plus added a test that tests resetNetworkFromParameters both with default parameters and a full set of parameters, with and without normalization.

maraboupy/MarabouNetworkNNet.py
alexopenu: Hi Guy, 

I think I've addressed all the points that you've raised, can you please have another quick look?

maraboupy/MarabouNetworkNNet.py
```diff
@@ -69,6 +209,10 @@ def __init__ (self, filename, normalize=False):
         # Set the number of variables
         self.numVars = self.numberOfVariables()
 
+    # def getMarabouQuery(self):
```
guykatzz, line 212: Please remove commented-out code from the PR

maraboupy/MarabouNetworkNNet.py
```diff
@@ -141,18 +285,108 @@ def read_nnet(self, file_name):
...
+            # The fifth line specifies the minimum values each input can take
+            # The sixth line specifies the maximum values each input can take
+            #     Inputs passed to the network are truncated to be between this range
+            # The seventh line gives the mean value of each input and of all outputs
+            # The eighth line gives the range of each input and of all outputs
+            #     These two lines are used to map raw inputs to the 0 mean, unit range of the inputs and outputs
+            #     used during training
+            # The ninth line begins the network weights and biases
+            ####################
+            # f2.write("// The contents of this file are licensed under the Creative Commons\n")
```
guykatzz, line 326: Likewise, please remove commented-out code

guykatzz:
Apart from a couple of comments about removing commented-out code from the PR, I don't have additional remarks. 
Please note that the Travis integration tests are failing - this also needs to be addressed before the code can be merged.
(And also we should wait for a LGTM from Kyle.)

maraboupy/MarabouNetworkNNet.py
alexopenu: actually, I wasn't sure whether these lines should or should not be in the header of a file created by writing an nnet into a file, what do you think?

maraboupy/MarabouNetworkNNet.py
alexopenu: Done. Sorry, this method moved to the parent class, and I forgot to remove the commented out lines from here. 

Indeed, I have been having build issues with a new test that I've added (still figuring out how certain things work in pytest). Hopefully the last commit fixes the issue. Thanks for the comments!

maraboupy/MarabouNetworkNNet.py
alexopenu: I have now resolved some conflicts that resulted from #317 , and also adjusted docustrings for the new methods to match the new style. 

In addition, I have noticed that there are two ways of creating a Marabou input query from the python interface: one by calling the getInputQuery() method of MarabouNetwork class, and the other one by asking marabou to create one directly (via the interface with the marabou InputQuery class in MarabouCore). This worried me a little, so I added a test to test_query.py that compares these two input queries (by saving them both to temporary files and comparing these files). 

@kjulian3, can you please have another look? I prefer to merge this PR soon, before new conflicts arise. :) 

Thanks!

kjulian3:
I only have two more general comments. First, some of the functions you've added don't have any tests associated with them, as seen in the [codecov report](https://codecov.io/gh/NeuralNetworkVerification/Marabou/compare/5fa9abf12a5e0d40ee6bed10d567548a553192ad...73e9b5a3bc1942f2735ca14430c9eb7039733013/diff). I assume that these functions are useful for your work and are important to have. Is it possible to add a test for them?

kjulian3:
Second, some function docstrings comments are not quite right. They should follow [this syntax](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) as closely as possible, paying particular attention to empty lines between the different "sections", and how class Attributes should be listed (no need to separate by where the attributes get defined, or to give the parent class attributes). I'll add some comments on different docstrings that I notice are not quite right. 

This is important because these docstrings get automatically processed and used to build API documentation, which can be found [here](https://neuralnetworkverification.github.io/Marabou). This website is automatically updated every time the master branch is updated, but if the docstring formatting is not correct, the generated documentation will be messed up too. This is a new thing I've been working on, so I just set it up and am still working on creating a developer's guide to explain the docstring syntax.

kjulian3:
If you want to check that the documentation builds correctly, you can build locally by following the instructions in the README in the maraboupy/docs folder. To do that, you'll have to update your pull request branch with changes made to NeuralNetworkVerification's master branch, which you can do with:
```
git fetch upstream
git checkout marabou-nnet-extensions
git rebase upstream/master
git push origin marabou-nnet-extensions -f
```

@guykatzz is this the way you would recommend to update a pull request with changes made to master?

maraboupy/MarabouNetworkNNet.py
```diff
@@ -13,40 +13,180 @@
...
-        inputMaximums (list of float): Maximum value for each input
-        inputMeans (list of floats): Mean value for each input
-        inputRanges (list of floats): Range for each input
-        outputMean (float): Mean value of outputs
-        outputRange (float): Range of output values
-        weights (list): Network weight matrices, where the outer index corresponds to layer number
-        biases (list): Network bias vectors, where the outer index corresponds to layer number
-     """
-    def __init__ (self, filename, normalize=False):
+    """ Abstract class representing a Marabou network from an NNet file
```
kjulian3, line 26: Remove "Abstract", since this is not an abstract class. There should be an "Attributes:" section that lists attributes of this class (not the ones from parent class). The attributes list should not be separated by where they were initialized. It should follow the same format as that of the original MarabouNetworkNNet docstring. Also, for consistency, the description for f_variables should begin with capital "List" instead of "list".

kjulian3: Also, there should be an Args section here, and the docstring for \_\_init\_\_ function should be omitted.

The Args section in the original docstring has a couple of errors. The filename argument doesn't have a type, and the colon should come after the type. So it should look something like:

filename (str): path to ...
normalize (bool): True if ...

It would be great to fix my mistake here.

maraboupy/MarabouNetworkNNet.py
```diff
@@ -13,40 +13,180 @@
...
+            lowerBounds (Dict[int, float]): Lower bounds of variables
+            upperBounds (Dict[int, float]): Upper bounds of variables
+            inputVars (list of numpy arrays): Input variables
+            outputVars (numpy array): Output variables
+
+    """
+
+
+    def __init__(self, filename='', normalize=False):
+        """Constructs a MarabouNetworkNNet object from an .nnet file
```
kjulian3, line 65: No docstring needed for __init__, since it is already covered by the class docstring.

maraboupy/MarabouNetworkNNet.py
```diff
@@ -13,40 +13,180 @@
...
+        self.inputRanges = inputRanges
+        self.outputMean = outputMean
+        self.outputRange = outputRange
+        self.weights = weights
+        self.biases = biases
+
+        self.computeNetworkAttributes()
+        # TODO (necessary? desirable?): create a new filename for the network
+
+    def computeNetworkAttributes(self):
```
kjulian3, line 189: Short docstring needed here. We should add the ":meta private:" at the end of the docstring too, which makes the method a "private" method and hides the method from the API documentation.

maraboupy/MarabouNetworkNNet.py
```diff
@@ -141,18 +281,108 @@ def read_nnet(self, file_name):
...
 
+
+
+    def writeNNet(self, file_name: str):
+        """Write network data into the .nnet file format
+
+        Args:
+            file_name (str): File where the network will be written
+
+        :meta private:
```
kjulian3, line 297: The ":meta private:" line makes the method private so that the method won't appear in documentation, which is for methods that ideally a general user wouldn't use. However, it seems like this might be a useful for a user to have available so they can write their own NNet file? If so, I'd recommend removing the ":meta private: line.

maraboupy/MarabouNetworkNNet.py
```diff
@@ -218,11 +447,204 @@ def nodeTo_f(self, layer, node):
             return node
         else:
             offset = self.layerSizes[0]
-            offset += sum([x*2 for x in self.layerSizes[1:layer]])
+            offset += sum([x * 2 for x in self.layerSizes[1:layer]])
             offset += self.layerSizes[layer]
 
             return offset + node
 
+    def getVariable(self, layer, node, f=True):
```
kjulian3, line 455: Docstrings needs to have an empty line before "Args:" and before "Returns:". Also, "Return:" should be changed to "Returns:" Also, the first line should be a quick description, followed by an empty line, and then a longer description can be written in a separate section. So I'd recommend an empty line between """Returns variable ... and If f=True lines.

Also, if these functions are private and only useful for creating the MarabouNetworkNNet class, the :meta private: line should be added to the end of the docstring.

It would also be great to add some sort of quick description of the arguments, using the format:
variable (type): Description

These comments applies to the other functions that have been created.

maraboupy/MarabouNetworkNNet.py
```diff
@@ -218,11 +447,204 @@ def nodeTo_f(self, layer, node):
...
+            layer (int)
+            f (bool)
+        Returns:
+            tuple of two list of floats (1st is lower bounds)
+        """
+        return self.getLowerBoundsForLayer(layer, f), self.getUpperBoundsForLayer(layer, f)
+
+
+    def numberOfVariables(self):
+        return self.layerSizes[0] + 2 * sum(self.layerSizes[1:-1]) + 1 * self.layerSizes[-1]
```
kjulian3, line 540: Docstrings are needed for these functions.

maraboupy/MarabouNetworkNNet.py
```diff
@@ -218,11 +447,204 @@ def nodeTo_f(self, layer, node):
...
+
+    def getInputMinimum(self, inputVar):
+        return self.inputMinimums[inputVar]
+
+    def getInputMaximum(self, inputVar):
+        return self.inputMaximums[inputVar]
+
+
+    def evaluateWithoutMarabou(self, inputValues):
+        """ Evaluate network directly (without Marabou) at a given point
```
kjulian3, line 550: Empty lines needed before Args: and Returns:

maraboupy/MarabouNetworkNNet.py
```diff
@@ -218,11 +447,204 @@ def nodeTo_f(self, layer, node):
...
+        Returns:
+            (np array): output of the network
+        """
+
+        return self.evaluateNNet(inputValues.flatten().tolist(), normalize_inputs=self.normalize,
+                                 normalize_outputs=self.normalize)
+
+
+    def evaluateNNet(self, inputs, first_layer = 0, last_layer=-1, normalize_inputs=False,
+                     normalize_outputs=False, activate_output_layer=False):
```
kjulian3, line 562: In Args: list, colon should appear after (type). So the arguments should all look like

variable (type): Description

maraboupy/MarabouNetworkNNet.py
```diff
@@ -218,11 +447,204 @@ def nodeTo_f(self, layer, node):
...
+                output_mean = means[layer - 1]
+                output_range = ranges[layer - 1]
+
+            for i in range(output_size):
+                outputs[i] = outputs[i] * output_range + output_mean
+
+        return outputs
+
+
+    def createRandomInputsForNetwork(self):
```
kjulian3, line 636: Docstring here

maraboupy/MarabouNetworkNNet.py
alexopenu: @kjulian3, thanks for the helpful and detailed comments! I am now working on making the changes, and I wanted to ask your advice on something. I am not sure how to treat  attributes (members, not methods) of the class that are not initialized in __init__(): e.g., b_variables in this case. On the one hand, it makes sense to me to include them in the docustring (they were omitted previously). On the other hand, it also makes sense to me to separate them from the other members, if nothing else, as a warning to the developer (after all, documentation should be helpful to us as well as the users -- at least, I believe that it would have been helpful for me if all the attributes had been listed in the docustring when I had started making changes, and it would have also been [mildly] helpful if there had been an indication to where things first "appear"). This is why originally I indicated explicitly that they are initiated elsewhere). Any suggestions/thoughts?

kjulian3: It looks like *init*() calls *self.computeNetworkAttributes*(), which call *self.variableRanges*(), which sets the attributes b_variables and f_variables. This happens any time init is called and filename is not empty. If filename is empty, then *self.clearNetwork*() is used instead, so I think b_variables and f_variables should be initialized there to empty lists. 

You could also call clearNetwork at the beginning of init regardless of filename, which would explicitly set all attributes to the default empty/0 values. Then if filename is not empty, the read_nnet and computeNetworkAttributes can be called to update the nominal values.

If finding where a variable is initialized is tricky, we could add "Initialized in self.variablesRanges()" to the attribute's description, for example.

alexopenu: Thanks, this sounds good.

maraboupy/MarabouNetworkNNet.py
alexopenu: @kjulian3, would you mind having another look (hopefully one last time)? Thanks!

maraboupy/test/test_nnet_extensions.py
```diff
@@ -0,0 +1,231 @@
+
+from maraboupy.MarabouNetworkNNet import *
+from maraboupy import Marabou
+
+TOL = 1e-8
+
+network_filename = "./resources/nnet/acasxu/ACASXU_experimental_v2a_1_9.nnet"
+
+layer = 2
+
```
kjulian3, line 10: It's convention to write global variable constants using all uppercase letters (like "TOL" instead of "tol"), and this is the convention used in other test files. This also makes reading  and editing the code easier in the future.

Also, defining the path to the nnet file like this means you can only run the tests from the Marabou root folder, and you will get errors if you run the tests from a different folder. This can be remedied with os.path.join() and os.path.dirname(\_\_file\_\_). Please see the other test files as examples.

There are also a lot of extra empty lines in this file. Generally, a single empty line can be used to separate blocks of code or methods, and an empty line usually comes before a comment line, but there should only be a single empty line. The file should end with a single empty line, and the file should begin with a comment or line of code (not an empty line).

maraboupy/test/test_query.py
```diff
@@ -100,6 +102,28 @@ def test_to_query(tmpdir):
...
+
+    ipq1_filename = tmpdir.mkdir("query").join("query1.txt").strpath
+    ipq2_filename = tmpdir.join("query2.txt").strpath
+
+    MarabouCore.saveQuery(ipq1, ipq1_filename)
+    MarabouCore.saveQuery(ipq1, ipq2_filename)
+    diff = call(['diff', ipq1_filename, ipq2_filename])
+    assert not diff
+
+
```
kjulian3, line 126: Only one empty line should separate the assert and load_onnx_network

maraboupy/MarabouNetworkNNet.py
```diff
@@ -141,18 +270,106 @@ def read_nnet(self, file_name):
...
             self.biases = biases
 
-            # Convert input bounds from their original values to normalized values if normalization is not being incorporated
             if not self.normalize:
                 for i in range(self.inputSize):
                     self.inputMinimums[i] = (self.inputMinimums[i] - self.inputMeans[i]) / self.inputRanges[i]
                     self.inputMaximums[i] = (self.inputMaximums[i] - self.inputMeans[i]) / self.inputRanges[i]
 
+
+
```
kjulian3, line 279: There are a few instances of double blank lines in this file that should be reduced to a single blank line to be consistent with the rest of Marabou files. I know it seems minor and picky, but I just want Marabou to have a clean, consistent, professional style. Generally it's best to follow the style of the existing code, or to follow the official python style guide: https://www.python.org/dev/peps/pep-0008/

alexopenu: Fixed. 

I personally prefer double-blank lines around class members (I think it makes the code much more readable), but if the rest of maraboupy follows the PEP-08 convention, I am happy to go along -- I completely agree that a bit project it is important to have uniformity and consistency.

maraboupy/test/test_query.py
alexopenu: Actually, even PEP-08 suggests surrounding top-level functions by two blank lines (and my editor, that I have now set to warn me on convention violations, complains when I try I remove this blank, so I've left it as is.

maraboupy/test/test_nnet_extensions.py
alexopenu: Changes made.

maraboupy/MarabouCore.cpp
```diff
@@ -270,7 +270,7 @@ PYBIND11_MODULE(MarabouCore, m) {
         .def("setUpperBound", &InputQuery::setUpperBound)
         .def("setLowerBound", &InputQuery::setLowerBound)
         .def("getUpperBound", &InputQuery::getUpperBound)
-        .def("getLowerBound", &InputQuery::getLowerBound)
+                .def("getLowerBound", &InputQuery::getLowerBound)
```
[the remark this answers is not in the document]
alexopenu: ooops, you are right, not sure what happened here...

maraboupy/MarabouNetworkNNet.py
```diff
@@ -163,66 +377,305 @@ def variableRanges(self):
...
             layer (int): Layer number
-            node (int): Node number
+            node (int): Node index within layer
 
         Returns:
             (int)
-        
-        :meta private:
+
+        :meta-private:
```
kjulian3, line 404: When using the meta private tag, there shouldn't be a hyphen between meta and private. It should be just ":meta private:", otherwise the auto-documentation will look messed up.

I think this is the last comment I have. So before we merge we'll need to update this PR with changes made to the Marabou master branch to make sure travis builds correctly and the documentation looks correct. @guykatzz did you have any additional comments?

kjulian3: The meta-private needs to be changed in nodeTo_f too.

alexopenu: Done. Thanks!

@guykatzz, any more comments? Can we merge this?

maraboupy/MarabouNetworkNNet.py
guykatzz: I don't care much either way - but we should either uncomment or remove

maraboupy/test/test_query.py
guykatzz: I suggest adhering to the convention used in the other files in the repo

alexopenu:
> Minor comments. Otherwise, if Kyle is happy, I think we can merge.

@kjulian3, are you happy? :)

I've made minor changes in the new test file, making sure that it follows the same style as the rest.

maraboupy/MarabouNetworkNNet.py
```diff
@@ -163,66 +375,305 @@ def variableRanges(self):
...
+        Evaluates the network on input inputs between layer first_layer (input layer by default)
+        and layer last_layer (output layer by default)
+
+        Args:
+            inputs: (list of float):         Network inputs to be evaluated
+            first_layer:  (int):              The initial layer of the evaluation
+            last_layer: (int):                The last layer of the evaluation
+            normalize_inputs: (bool):         If True and first_layer==0, normalization of inputs is performed
+            normalize_outputs: (bool):        If True, normalization of output is undone
+            activate_output_layer: (bool):    If True, the last layer is activated, otherwise it is not.
```
kjulian3, line 597: Please remove the colon after parameter names. There should only be a colon after the (parameter type).

Docstring syntax error like this can be found by building the documentation locally, following the instructions in the maraboupy/docs folder. I did this just now and saw that the parameter formatting for this function was a little off. Once this is fixed and we merge into master, [this website](https://neuralnetworkverification.github.io/Marabou/API/2_MarabouNetworkNNet.html) will automatically be updated with the new functions, class attributes, and with your name in the contributor section, so it's best practice to build the documentation locally and make sure that the documentation looks consistent. :) 

Everything else looks good, so I'm happy to merge once this is fixed. Thanks for your work!

alexopenu: Thank you. Sorry this review must have involved quite a bit of work..

PR #311, opened 2020-06-28
remove maraboupy README, simplify python part in main README

### LINKED ISSUE #307: The maraboupy directory does not have a Makefile
The [installation instructions for maraboupy](https://github.com/NeuralNetworkVerification/Marabou/tree/master/maraboupy#installation) say to call `make` from the `maraboupy` directory. However, this directory does not contain a Makefile.

yuvaljacoby:
fix #307


kjulian3:
Is build_python_x86.sh still up to date, or should this be removed? There's no mention of it in the readme.

yuvaljacoby:
> Is build_python_x86.sh still up to date, or should this be removed? There's no mention of it in the readme.

good point!
It's an example (maybe not a very good one) for installing a python32 bit interface, there is a warning when building that points to there.
added a reference from the README

PR #312, opened 2020-06-28
Query dump

guykatzz:


guykatzz:
@anwu1219 
Hi Andrew,
We didn't have this functionality before, right? I saw that the python interface has save/load functionality, but I think the C++ interface only had load functionality.

anwu1219:
> @anwu1219
> Hi Andrew,
> We didn't have this functionality before, right? I saw that the python interface has save/load functionality, but I think the C++ interface only had load functionality.

I don't believe. This is very nice to have!

PR #326, opened 2020-07-17
Fix evaluateWithMarabou when system is UNSAT

vbCrLf:
Instead of throwing KeyError exception when given an UNSAT system, evaluateWithMarabou (and so evaluate as well) now returns None.


kjulian3:
Looks good to me. I added a test for these changes, so let me know if the test I added looks good to you.

vbCrLf:
Looks good Julian :)

PR #327, opened 2020-07-17
Add abs constraint support to maraboupy

vbCrLf:


kjulian3:
Everything here looks setup and formatted correctly, but how can this absolute value constraint be used as written? The tensorflow/onnx input parsers haven't been changed, so one would need to manually create a neural network without using an input parser, or one would need to manually edit a network parsed from a file and add new "forward" variables for the abs constraint. We should write a test case for this new constraint, but I'm not sure how to do it without making edits to the tensorflow/onnx parsers to incorporate abs operations.

vbCrLf:
I'll explain my use case.
I load an NNet and manipulate it in some ways. It is easier for me to use the MarabouNetwork class and not the query directly for three reasons:
1. I use the notions of neurons in my manipulation
2. It is easier to manipulate a pure Python class (oppose to InputQuery)
3. It has convenient methods such as evaluate. 

MarabouNetwork is a general class for networks, I thought it might be okay that it will have features not used by a specific subclass (such as NNet or TF) but left for future/special networks types.

At the end it is a style decision, so whatever you say is okay. We can only implement it in MarabouCore and I'll subclass MarabouNetwork with support for Abs for my personal use.

kjulian3:
I think what you've done with MarabouNetwork is perfect and follows exactly how max and relu constraints are handled. I just want to add a test case for the abs constraint to make sure everything runs correctly, and I'm not used to manipulating the MarabouNetwork class directly as you do. Are you able to write a simple test case that uses the abs constraint, something like what I did in your other pr? 

In the future I can add the support for abs operations to the TF/ONNX parsers that will use what you've created here. If it's difficult to write a test case that manipulates MarabouNetwork, I can add the TF/ONNX abs operations to this pr and write a test case that uses those parsers.

vbCrLf:
Sorry, I misunderstood you.
Added a test. It is currently fails, but I don't know if it fails due to problems in the test or in the abs implementation (I'm having issues with using abs constraint right now, maybe the fixes Guy is pushing right now may help here).

maraboupy/test/test_network.py
```diff
@@ -0,0 +1,58 @@
...
+        [abs(-0.02158248), -0.01885345, abs(-0.01892334), -0.01892597, -0.01893113, abs(testInputs[1][0])],
+        [abs(0.05990158), 0.05273383, abs(0.10029709), 0.01883183, 0.10521622, abs(testInputs[2][0])]
+    ]
+
+    network = loadNetwork(filename)
+
+    # Replace two output variables with their's absolute value
+    for out in [0, 2]:
+        abs_out = network.getNewVariable()
+        network.outputVars[0][out] = abs_out
```
guykatzz, line 34: I think switching lines 34 and 35 fixed the problem. Please check and let me know.

vbCrLf:
This is embarrassing :) Now it works.

kjulian3:
Travis is failing for Windows and Linux builds. I tried reproducing the issue, and I get errors when using the absolute value constraint when the "b" variable is an input variable. For a windows machine, the error output is:  "Windows fatal exception: code 0xc0000374", which indicates heap corruption.

Looking at the log, Marabou seems to run and find  a SAT point, so there must be an error at the end of the solve method that causes this problem. I was able to run the test successfully on my Linux machine, so I'm not sure why Travis fails for Linux. 

@guykatzz do you have any ideas what could be causing the issue?

guykatzz:
I fixed the Linux build that was failing. One Linux build is configured for Debug and one for Release; and the Debug one was failing because of a too-strong assumption, which I removed.

Will try to look into the Windows error next.

guykatzz:
You can go ahead

PR #331, opened 2020-07-22
Fix file redirection

vbCrLf:
Output redirection does not get reverted on all cases.
This causes Maraboupy to crash on `output=redirectOutputToFile(redirect);` on next `solve` invocation.
Because there are many return statements in this function I added a cleanup label at the end of the function.

### DIFF
--- a/maraboupy/MarabouCore.cpp
+++ b/maraboupy/MarabouCore.cpp
@@ -150,7 +150,10 @@ std::pair<std::map<int, double>, Statistics> solve(InputQuery &inputQuery, Marab
         Engine engine;
         engine.setVerbosity(verbosity);
 
-        if(!engine.processInputQuery(inputQuery)) return std::make_pair(ret, *(engine.getStatistics()));
+        if(!engine.processInputQuery(inputQuery)) {
+            retStats = *(engine.getStatistics());
+            goto cleanup;
+        }
         if ( dnc )
         {
             unsigned initialDivides = options._initialDivides;
@@ -177,14 +180,17 @@ std::pair<std::map<int, double>, Statistics> solve(InputQuery &inputQuery, Marab
             {
                 retStats = Statistics();
                 retStats.timeout();
-                return std::make_pair( ret, retStats );
             }
             default:
-                return std::make_pair( ret, Statistics() ); // TODO: meaningful DnCStatistics
+                retStats = Statistics(); // TODO: meaningful DnCStatistics
+                goto cleanup;
             }
         } else
         {
-            if(!engine.solve(timeoutInSeconds)) return std::make_pair(ret, *(engine.getStatistics()));
+            if(!engine.solve(timeoutInSeconds)) {
+                retStats = *(engine.getStatistics());
+                goto cleanup;
+            }
 
             if (engine.getExitCode() == Engine::SAT)
                 engine.extractSolution(inputQuery);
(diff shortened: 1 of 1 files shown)


vbCrLf:
Bug reproduction:
`MarabouCore.solve(MarabouCore.loadQuery("query.ipq"), MarabouCore.Options())`

`query.ipq`:
`20
9
5
11
4
5
0,0
1,1
2,2
3,3
4,4
5
0,15
1,16
2,17
3,18
4,19
0,0.0000000000
1,0.0000000000
2,0.0000000000
3,0.0000000000
4,0.0000000000
11,0.0000000000
12,0.0000000000
13,0.0000000000
14,0.0000000000
0,0.0000000000
1,0.0000000000
2,0.0000000000
3,0.0000000000
4,0.0000000000
0,0,-0.2276300000,5,-1.0000000000,0,0.0540062000,1,-2.6109200000,2,-0.1800270000,3,0.2421940000,4,0.1414070000
1,0,0.1887620000,6,-1.0000000000,0,-1.1237400000,1,0.0263619000,2,-0.0091792900,3,0.0556230000,4,-0.3276350000
2,0,-0.0534094000,7,-1.0000000000,0,0.1960190000,1,0.2421590000,2,0.6384520000,3,-0.4782650000,4,0.1425770000
3,0,0.3778610000,8,-1.0000000000,0,-1.6301500000,1,-0.0344447000,2,-0.0060549200,3,0.0112076000,4,-0.0104997000
4,0,0.0812531000,9,-1.0000000000,0,-0.3551330000,1,0.5659690000,2,0.2282670000,3,0.1773420000,4,-0.2080780000
5,0,-0.0835500000,15,-1.0000000000,10,-0.1842020000,11,0.0343834000,12,-0.1151420000,13,0.1847530000,14,-0.1371480000
6,0,0.0213261000,16,-1.0000000000,10,0.0171639000,11,-0.0344739000,12,-0.0086931100,13,0.0381665000,14,0.0214070000
7,0,-0.7562910000,17,-1.0000000000,10,0.0698282000,11,-0.0040891100,12,-0.2779970000,13,0.2587900000,14,0.1558200000
8,0,1.2508200000,18,-1.0000000000,10,0.7948370000,11,-0.3629800000,12,-0.1719110000,13,0.0975946000,14,0.0271402000
9,0,-0.1784690000,19,-1.0000000000,10,-0.0585128000,11,0.0070839400,12,0.2681160000,13,-0.0065952100,14,0.3158060000
10,0,0.0000000000,5,1.0000000000,10,-1.0000000000
0,relu,11,6
1,relu,12,7
2,relu,13,8
3,relu,14,9`

And two other related thoughts:
1. I think the `filename` should be off by default. (log is usually not needed)
2. I find myself patching the code a lot adding an option to save the evaluation query. I think it might be useful to have a "query_filename" parameter which saves the query to a file for later debugging.
If you agree on these points I can add a commit which implements these changes.

Ori

guykatzz:
For your comment #2, there's a C++ option 

https://github.com/NeuralNetworkVerification/Marabou/blob/c732d566bb1903db96644e0d35f2b49de404748e/src/configuration/OptionParser.cpp#L60

I think it's not exposed directly in the Python interface, but you have this function here which does the same:

https://github.com/NeuralNetworkVerification/Marabou/blob/c732d566bb1903db96644e0d35f2b49de404748e/maraboupy/MarabouCore.cpp#L209

vbCrLf:
@guykatzz yeah, the issue is that the query is created inside evaluateWithMarabou but does not get returned (or saved). I think we should add an option to return the query for later inspection.

guykatzz:
The C++ option will dump the query, no matter which Python interface function it is created through. We can just expose it to the Python side.

Issue #330, opened 2020-07-22
evaluateWithMarabou doesn't work with DNC


vbCrLf:
```
import numpy as np
from maraboupy import Marabou

NNET_FILE_NAME = "../Marabou/src/input_parsers/acas_example/ACASXU_run2a_1_1_tiny_2.nnet"
OPT = Marabou.createOptions(verbosity=100)
OPT_DNC = Marabou.createOptions(verbosity=100, dnc=True, numWorkers=8)

nn = Marabou.read_nnet(NNET_FILE_NAME)

print("Works")
nn.evaluate([[0, 1, 2, 3, 4, 5, 6, 7, 8 ,9]], options=OPT)

print("Doesn't work")
nn.evaluate([[0, 1, 2, 3, 4, 5, 6, 7, 8 ,9]], options=OPT_DNC)
```

`evaluateWithMarabou` doesn't work when using DNC. The code above reproduces the problem.
Using DNC for evaluation is an overkill, but it is easy to accidentally do when provide the same options for every call to Marabou, and it took me some time to find the problem.
I think it is enough to raise an exception when running evaluate with DNC, no need to fix it.

kjulian3:
@anwu1219 can you take a look at this? It looks like evaluating with DNC mode returns only the input and output variable assignments, and the output variable numbers have changed from 105-109 to 5-9. Maybe because other variables were eliminated during pre-processing? MarabouCore in DNC mode uses the dncManager's getSolution method, which is a bit different than how non-DNC mode extracts the solution.

Also, DNC mode in maraboupy is currently not tested. It would be great to add a test file for DNC mode and include this script as one of the tests.

anwu1219:
Hi, thanks for bringing up this issue! I created a PR(https://github.com/NeuralNetworkVerification/Marabou/pull/333) to fix this issue. Could you please try and see if it fixes your issue.

Andrew

anwu1219:
> @anwu1219 can you take a look at this? It looks like evaluating with DNC mode returns only the input and output variable assignments, and the output variable numbers have changed from 105-109 to 5-9. Maybe because other variables were eliminated during pre-processing? MarabouCore in DNC mode uses the dncManager's getSolution method, which is a bit different than how non-DNC mode extracts the solution.
> 
> Also, DNC mode in maraboupy is currently not tested. It would be great to add a test file for DNC mode and include this script as one of the tests.

[...]

PR #333, opened 2020-07-24
fix bug in dnc evaluate

anwu1219:
Fixing issue https://github.com/NeuralNetworkVerification/Marabou/issues/330. 
The issue is that when some worker finds a SAT instance and returns a mapping from variable indices to assignments, it is using variable indices after preprocessing. This is because the worker itself does not preprocess the query but directly take in the preprocessed query, so it does not store the mapping from original variable index to the new index.
This information is stored in DnCManager._baseEngine. Therefore, we copy the tableau state of DnCManager._engineWithSATAssignment into DnCManager._baseEngine and extract solutions from the latter object, instead of the former object.

### DIFF
--- a/maraboupy/MarabouCore.cpp
+++ b/maraboupy/MarabouCore.cpp
@@ -174,7 +174,7 @@ std::pair<std::map<int, double>, Statistics> solve(InputQuery &inputQuery, Marab
             case DnCManager::SAT:
             {
                 retStats = Statistics();
-                dncManager->getSolution( ret );
+                dncManager->getSolution( ret, inputQuery );
                 break;
             }
             case DnCManager::TIMEOUT:
--- a/maraboupy/test/test_dnc.py
+++ b/maraboupy/test/test_dnc.py
@@ -1,0 +1,90 @@
+# Tests Marabou's DNC option
+import pytest
+from .. import Marabou
+import os
+import numpy as np
+
+# Global settings
+OPT = Marabou.createOptions(verbosity=0, dnc=True, numWorkers=2) # Turn off printing, turn on DNC with two workers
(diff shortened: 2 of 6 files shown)


kjulian3:
We need to add tests for DNC mode in Maraboupy. I tried adding one just now, and ran into an unexpected error. Here's the test I was trying to run:

```
# Tests Marabou's DNC option
import pytest
from .. import Marabou
import os
import numpy as np

# Global settings
OPT = Marabou.createOptions(verbosity=2, dnc=True, numWorkers=2) # Turn off printing, turn on DNC with two workers
TOL = 1e-6                                                       # Set tolerance for checking Marabou evaluations
NETWORK_FOLDER = "../../resources/nnet/acasxu"                   # Folder for test networks

def test_dnc_unsat():
    """
    Test the 1,1 experimental ACAS Xu network. 
    Evaluate a simple local robustness property for a network with DNC mode activated
    """
    filename =  "ACASXU_experimental_v2a_1_1.nnet"
    filename = os.path.join(os.path.dirname(__file__), NETWORK_FOLDER, filename)
    network = Marabou.read_nnet(filename)
    
    # Local robustness input constraint
    centerPoint = [-0.2454504737724233, -0.4774648292756546, 0.0, -0.3181818181818182, 0.0]
    for var, val in zip(network.inputVars[0], centerPoint):
        network.setLowerBound(var, val - 0.002)
        network.setUpperBound(var, val + 0.002)

    # Solve with no output constraint
    # Should return SAT, and vals should be a populated dictionary
    vals, stats = network.solve(options = OPT, filename = "", verbose=True)
    assert len(vals) > 0
```

I created a new file with this code in it (maraboupy/test/test_dnc.py), and calling pytest via ```pytest test/test_dnc.py``` from the maraboupy directory will produce the Marabou Error, code 6.

PR #335, opened 2020-07-27
SBT bug fix

guykatzz:
A bug that occurs when a weighted sum layer has multiple layers feeding into it


vbCrLf:
The PR fixes the issue I had.

PR #341, opened 2020-07-31
Add a GlobalConfiguration option to not perform preprocessing

castrong:
Added a NONE option to the enum MILPSolverBoundTighteningType in GlobalConfiguration.h. In Engine.cpp, in the case where it is NONE no MILP tightening will be performed.


src/engine/Engine.cpp
```diff
@@ -1129,6 +1129,8 @@ void Engine::performMILPSolverBoundedTightening()
         case GlobalConfiguration::MILP_ENCODING_INCREMENTAL:
             _networkLevelReasoner->MILPPropagation();
             break;
+        case GlobalConfiguration::NONE:
+            break;
```
guykatzz, line 1133: I suggest to have a "return" instead of a "break", to save time on the following instructions

src/configuration/GlobalConfiguration.h
```diff
@@ -172,6 +172,8 @@ class GlobalConfiguration
         // more expensive to solve
         MILP_ENCODING = 2,
         MILP_ENCODING_INCREMENTAL = 3,
+				// Option to have no MILP bouond tightening performed
```
guykatzz, line 175: Typo, + funny indentation spacing

PR #342, opened 2020-08-01
Fix FP comparison in ConstraintBoundTightener to use FloatUtils

AleksandarZeljic:


guykatzz:
Hi Aleks,
I know this point is inconsistent in the code, but for bound tightening it's generally okay to use < and >, and not FloatUtils' versions. FloatUtils calls take up a nice portion of our runtime, so it's best to avoid them when possible.

AleksandarZeljic:
I see, sounds good. I'll close the PR.

PR #344, opened 2020-08-09
Guy Amir - added addSignConstraint to Maraboupy

guyam2:
same as the addReluConstraint() function


guykatzz:
Looks good to me; I'll just let @kjulian3 confirm that the new documentation guidelines are met.

kjulian3:
This looks good so far, but we'd probably want to make some changes to MarabouNetwork.py to expose the sign constraint to the network class and make it easier for users to access, and we should add some tests to make sure the sign constraint works correctly, as was done for #327. This could be done in a separate PR if needed. What's done here though looks correct to me

guykatzz:
Thanks Kyle. Our next step will indeed be to expose the sign constraint, probably through the TF interface, and we will make sure to test it.

PR #349, opened 2020-08-11
Reuse GurobiWrapper in LPFormulator

anwu1219:
Let LPFormulator own a GurobiWrapper, which can be reuse.

### DIFF
--- a/src/common/GurobiWrapper.cpp
+++ b/src/common/GurobiWrapper.cpp
@@ -16,6 +16,7 @@
 #ifdef ENABLE_GUROBI
 
 #include "Debug.h"
+#include "GlobalConfiguration.h"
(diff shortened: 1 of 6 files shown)


guykatzz:
LGTM, except lets make the Gurobi thread number a configurable parameter.
Also, can you compare performance to the previous version, to make sure nothing breaks?

anwu1219:
> LGTM, except lets make the Gurobi thread number a configurable parameter.
> Also, can you compare performance to the previous version, to make sure nothing breaks?

I made the thread number a configurable parameter.
I ran on 140 benchmarks, solved the same set of instances within 30 minutes.
master: 8678 (sat) + 7926 (unsat) sec
this branch: 8310 (sat) + 7609 (unsat) sec

PR #352, opened 2020-08-17
supporting sign activations in TF-formatted NN

guyam2:
code written by Guy Amir& Kyle, and tested by Guy Amir
all changes are in the Maraboupy interface


guyam2:
Updated pr # 353 from Kyle's user instead, includes the current sign update & an additional test I wrote for a NN with both sign & ReLU functions

PR #353, opened 2020-08-17
Sign activation for tensorflow

kjulian3:
This pr allows tensorflow models with the sign activation function (for binary networks) to be used with Marabou. A simple test was added that checks that Marabou and tensorflow give the same result when evaluating random test inputs. In addition, tensorflow parser now combines scalar multiplication and bias addition into one equation when possible to reduce the number of equations and variables.


guyam2:
@guykatzz - this is the updated one (I closed my pr # 352). Waiting for your approval

PR #357, opened 2020-08-20
Add MNIST benchmarks to resources

anwu1219:
Add some MNIST benchmarks to  ./resources
Augmented the regression tests with some targeted attack benchmarks


anwu1219:
On my local machine, regress0-1 time increased by 30 secs.

PR #361, opened 2020-08-25
Sparse constraint matrix analysis

guykatzz:
The dense matrix approach was becoming a bottleneck for larger networks


src/engine/Engine.cpp
```diff
@@ -773,7 +773,7 @@ void Engine::removeRedundantEquations( const double *constraintMatrix )
     analyzer->analyze( constraintMatrix, m, n );
 
     ENGINE_LOG( Stringf( "Number of redundant rows: %u out of %u",
-                  analyzer->getRedundantRows().size(), m ).ascii() );
+                         analyzer->getRedundantRows().size(), m ).ascii() );
```
anwu1219, line 776: remove spaces

src/engine/Engine.cpp
```diff
@@ -2105,11 +2105,3 @@ void Engine::setConstraintViolationThreshold( unsigned threshold )
 {
     _smtCore.setConstraintViolationThreshold( threshold );
 }
-
```
anwu1219, line 2108: Why are these lines removed?

guykatzz: Leftovers from when we were using Makefiles (and not CMake)

src/engine/Engine.cpp
guykatzz: The new spaces? It was not properly indented before, now it is directly below the "

src/engine/ConstraintMatrixAnalyzer.h
```diff
@@ -27,48 +28,57 @@ class ConstraintMatrixAnalyzer : public IConstraintMatrixAnalyzer
...
+
+    unsigned _pivotRow;
+    unsigned _pivotColumn;
+    double _pivotElement;
+
+    double *_workRow;
+    double *_workRow2;
+
     /*
       The i'th (permuted) column of the matrix is stored in memory
```
anwu1219, line 61: ith (permuted) row?

PR #365, opened 2020-08-31
improve addMaxConstratins

tagomaru:

This PR is going to improve _addMaxConstraints_ function.

### As-Is
It supports only the case of _max([A, B], C])._

### To-Be
It supports the case of _max([A, B, C], C)_ as well.

### Sample NNET:
```
// simple_model 
2,2,3,3,
2,2,3,
0,
-100000.0,-100000.0,
100000.0,100000.0,
0.0,0.0,0.0,
1.0,1.0,1.0,
1.0,1.0,
-1.0,-1.0,
0.0,
0.0,
2.0,-1.0,
-1.0,1.0,
1.0,-1.0,
0.0,
0.0,
0.0,
```
### Test for the above net
Command:
```
nnet_file_name = "./simple_model.nnet"
net = marabou.read_nnet(nnet_file_name)

net.addMaxConstraint(set(net.outputVars[0]), net.outputVars[0][0])
net.setLowerBound(net.outputVars[0][0], 2)
net.setUpperBound(net.inputVars[0][0], 1)

vals, stats = net.solve()
```

The result:
```
sat
input 0 = 1.0
input 1 = -0.0
output 0 = 2.0
output 1 = -1.0
output 2 = 1.0
```

image

### DIFF
--- a/src/engine/MaxConstraint.cpp
+++ b/src/engine/MaxConstraint.cpp
@@ -376,28 +376,42 @@ List<PiecewiseLinearConstraint::Fix> MaxConstraint::getSmartFixes( ITableau * ) 
 
 List<PiecewiseLinearCaseSplit> MaxConstraint::getCaseSplits() const
 {
-    if ( phaseFixed() )
+    if ( phaseFixed() && !_elements.exists( _f ) )
         throw MarabouError( MarabouError::REQUESTED_CASE_SPLITS_FROM_FIXED_CONSTRAINT );
 
     ASSERT(	_assignment.exists( _f ) );
 
     List<PiecewiseLinearCaseSplit> splits;
-    for ( unsigned element : _elements )
-	{
-        splits.append( getSplit( element ) );
-	}
+
+    if ( !_elements.exists( _f ) )
+    {
+        for ( unsigned element : _elements )
+        {
+            splits.append( getSplit( element ) );
+        }
+    }
+    else
+        // if elements includes _f, 
+        // this should return not a piecewise linear constraint but the linear constratint such as _f >= x1 /\ _f >= x2.
+        splits.append( getSplit( _f ) );
+
     return splits;
 }
 
 bool MaxConstraint::phaseFixed() const
 {
-    return _elements.size() == 1;
+    return _elements.size() == 1 || _elements.exists( _f );
 }
 
 PiecewiseLinearCaseSplit MaxConstraint::getValidCaseSplit() const
 {
     ASSERT( phaseFixed() );
-    return getSplit( *( _elements.begin() ) );
+    if ( !_elements.exists( _f ) )
+        return getSplit( *( _elements.begin() ) );
+    else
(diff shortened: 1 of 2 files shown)


src/engine/MaxConstraint.cpp
```diff
@@ -85,35 +85,39 @@ void MaxConstraint::restoreState( const PiecewiseLinearConstraint *state )
...
+    if ( !_elements.exists( _f ) )
+        tableau->registerToWatchVariable( this, _f );
 }
 
 void MaxConstraint::unregisterAsWatcher( ITableau *tableau )
 {
-    tableau->unregisterToWatchVariable( this, _f );
+    // Guy: I think there's a bug here, not related to the recent changes:
+    // the list of _elements can change, and so some of them might not
+    // get propertly unregistered.
```
tagomaru, line 99: Yes.
When element is removed, it should be unregistered.
But this does not have function which unregister just one element.

src/engine/MaxConstraint.cpp
```diff
@@ -85,35 +85,39 @@ void MaxConstraint::restoreState( const PiecewiseLinearConstraint *state )
...
+        tableau->unregisterToWatchVariable( this, _f );
 }
 
 void MaxConstraint::notifyVariableValue( unsigned variable, double value )
 {
+    /*
+      Guy: I'm not sure I understand this condition. If _f is an
+      element, what is the _maxIndex? Is it always _f? Is it the
+      highest element if that's not _f?
+    */
```
tagomaru, line 113: I assume:
If _f is not an element, _maxIndex is the highest element of _elements.
If _f is an element, maxIndex is the highest element of _elements including _f.

Namely, _maxIndex is always the highest element of elements.

src/engine/MaxConstraint.cpp
```diff
@@ -284,6 +288,8 @@ void MaxConstraint::resetMaxIndex()
     double maxValue = FloatUtils::negativeInfinity();
     _maxIndexSet = false;
 
+    // Guy: it looks to me like the max index should be set even if
+    // only _f is assigned, if _f is in _elements. What do you think?
```
tagomaru, line 292: Yes, I should have fix this as well and will do.

src/engine/MaxConstraint.cpp
guykatzz: Okay. Lets ignore this for now, and I'll open an issue so that we can take care of it later.

src/engine/MaxConstraint.cpp
guykatzz: Okay.

src/engine/MaxConstraint.cpp
guykatzz: Okay, so lets fix this before we merge the PR.

tagomaru: I fixed this.

guykatzz (approved):

PR #370, opened 2020-09-09
Interval splitting

anwu1219:
Add an input splitting strategy to the internal engine, where input splitting is used every K (GlobalConfiguration::INTERVAL_SPLITTING_FREQUENCY) split. The optimal value of K found so far on the ACAS benchmarks is 5.

Test on 180 ACAS properties, with lp preprocessing, 1hr timeout.

run | interval-split | relu-split
-- | -- | --
SAT | 39 | 39
UNSAT | 108 | 100
SAT-TIME | 8226 | 7197
UNSAT-TIME | 49770 | 45528


src/engine/DivideStrategy.h
```diff
@@ -22,6 +22,8 @@ enum class DivideStrategy
     Polarity = 0,      // Pick the ReLU with the polarity closest to 0 among the first K nodes
     EarliestReLU,  // Pick a ReLU that appears in the earliest layer
     ReLUViolation, // Pick the ReLU that has been violated for the most times
+    LargestInterval, // Pick the largest interval
```
guykatzz, line 25: So, it's really "largest interval every k steps, otherwise ReluViolation", right?

src/engine/Engine.cpp
```diff
@@ -47,6 +48,7 @@ Engine::Engine()
     , _verbosity( Options::get()->getInt( Options::VERBOSITY ) )
     , _lastNumVisitedStates( 0 )
     , _lastIterationWithProgress( 0 )
+    , _splittingStrategy( GlobalConfiguration::SPLITTING_HEURISTICS )
```
guykatzz, line 51: So, recently we've been treating mutable configuration options by moving them from GlobalConfig to Options. That way you can set it at will, and also access it without passing it around. I suggest to do that here, as well.

src/engine/PiecewiseLinearConstraint.cpp
```diff
@@ -17,7 +17,8 @@
 #include "Statistics.h"
 
 PiecewiseLinearConstraint::PiecewiseLinearConstraint()
-    : _constraintActive( true )
+    : _temporary( false )
```
guykatzz, line 20: This _temporary thing is a bit messy. Instead, lets pass the DisjunctionConstraint by value. It's very light weight, so no problem if it gets copied a couple of times.

src/engine/DivideStrategy.h
anwu1219: yes. Updated the documentation.

src/engine/Engine.cpp
anwu1219: Moved to Options

src/engine/PiecewiseLinearConstraint.cpp
anwu1219: Create a field variable in SmtCore to store this constraint.

src/engine/Engine.cpp
```diff
@@ -2028,20 +2038,67 @@ PiecewiseLinearConstraint *Engine::pickSplitPLConstraintBasedOnTopology()
...
 
-PiecewiseLinearConstraint *Engine::pickSplitPLConstraint( DivideStrategy strategy )
+PiecewiseLinearConstraint *Engine::pickSplitPLConstraintBasedOnIntervalWidth()
+{
+    // We push the first unfixed ReLU in the topology order to the _candidatePlConstraints
+    ENGINE_LOG( Stringf( "Using LargestInterval heuristics..." ).ascii() );
+
+    unsigned inputVariableWithLargestInterval = 0;
+    double largestIntervalSoFar = 0;
+    for ( const auto&variable : _preprocessedQuery.getInputVariables() )
```
guykatzz, line 2048: Add space before &

src/engine/Engine.cpp
```diff
@@ -2028,20 +2038,67 @@ PiecewiseLinearConstraint *Engine::pickSplitPLConstraintBasedOnTopology()
...
+                                             mid, Tightening::LB ) );
+
+        List<PiecewiseLinearCaseSplit> splits;
+        splits.append( s1 );
+        splits.append( s2 );
+        auto bisection = std::unique_ptr<DisjunctionConstraint>
+            ( new DisjunctionConstraint( splits ) );
+        PiecewiseLinearConstraint *bisectionPtr = bisection.get();
+        _smtCore.setDisjunctionForSplitting( std::move( bisection ) );
+        return bisectionPtr;
```
guykatzz, line 2080: I didn't get this part. We both return a pointer to the constraint, and store it in the SmtCore? Is this needed so that the SmtCore becomes the owner of the object?

Simpler to let the Engine stay the owner...

anwu1219: That makes sense. I made the Engine the owner

guykatzz (approved):

Issue #368, opened 2020-09-06
hidden layer properties


alexopenu:
1) I had the impression that hidden layer properties (i.e., bounds of the form 'h_i_j <= ..' in the property file) are currently supported by Marabou, but when trying to create an input query with the network ACASXU_experimental_v2a_2_7.nnet and the attached property, I get the following runtime error:

./resources/nnet/acasxu/ACASXU_experimental_v2a_2_7.nnet
Number of layers: 8. Input layer size: 5. Output layer size: 5. Number of ReLUs: 300
Total number of variables: 610
Property: new_input_property_acas2_7_prop4_level5_test1.txt
Traceback (most recent call last):

MarabouCore.createInputQuery(conjunction_ipq, conjunction_network, new_property_filename)
RuntimeError: Caught an unknown exception!

Am I missing something silly here?

2) A related question: by examining the Marabou property parser [here](https://github.com/NeuralNetworkVerification/Marabou/blob/master/src/input_parsers/PropertyParser.cpp#L141), it seems that only bounds on odd layers are supported; so in case of a simple feedforward network (e.g. ACAS networks) we are only allowing to express properties on layers of type "weighted sum", just like before (when expressions with variables of the form 'ws_i_j' were supported instead). 

Is this what we want?

[new_input_property_acas2_7_prop4_level5_test1.txt](https://github.com/NeuralNetworkVerification/Marabou/files/5180443/new_input_property_acas2_7_prop4_level5_test1.txt)

alexopenu:
OK, I think that I understand the issue. It seems that the two issues that I have raised are, in fact, related. The property that I am trying to use makes sense for a ReLU layer, but not for a weighted sum layer. I will try to see whether allowing expressing properties on all layers (and not just the odd layers) fixes the problem.

PR #371, opened 2020-09-10
Create nlr in python

alexopenu:
This pr addresses issue #368. Specifically, it includes three changes:

1. Construct Network Level Reasoner in MarabouCore.createInputQuery() function.
2. Catch InputParserError that is potentially thrown during the run of MarabouCore.createInputQuery() in order to avoid a run time error with an anonymous exception.
3. Allow expressing properties on all nlr levels, not just the odd ones (e..g, only levels of type "weighted sum"). This may appear more cumbersome to some users, but also allows more flexibility - at least I find it much more convenient.


maraboupy/MarabouCore.cpp
```diff
@@ -109,16 +110,29 @@ void addAbsConstraint(InputQuery& ipq, unsigned b, unsigned f){
...
+      {
+        printf( "Property: %s\n", propertyFilePathM.ascii() );
+        PropertyParser().parse( propertyFilePathM, inputQuery );
+      }
+    else
+      printf( "Property: None\n" );
+  }
+  catch(const InputParserError &e){
+        printf( "Caught an InputParserError. Code: %u. Message: %s\n", e.getCode(), e.getUserMessage() );
+        exit(1);
```
guykatzz, line 134: exit(1) is a very violent way to handle an exception. How are C++ exceptions handled in other areas of the Python code?

src/input_parsers/PropertyParser.cpp
```diff
@@ -138,7 +138,7 @@ void PropertyParser::processSingleLine( const String &line, InputQuery &inputQue
 
             auto subToken = subTokens.begin();
             ++subToken;
-            unsigned layerIndex = 2 * atoi( subToken->ascii() ) - 1;
```
guykatzz, line 141: Yes, this makes sense. It was originally put in to maintain compatibility with the old ws_ properties, now that WS+ReLU layers are stored as two separate layers. Lets just check with @anwu1219 that this doesn't break anything of his.

maraboupy/MarabouCore.cpp
alexopenu: You are right, common practice seems to do nothing, that is, return with whatever partial results one may have; so this is what I have switched to now (there is nothing to return, but the input query may have been partly populated by this point -- which is exactly what was happening in my original example).

src/input_parsers/PropertyParser.cpp
anwu1219: Sorry somehow just saw this message. I think this should be fine on my end. Maybe it'd be good to put the convention that we are treating ws and relu as two separated layer somewhere in the README.

anwu1219 (approved):

src/input_parsers/PropertyParser.cpp
alexopenu: Thanks, Andrew.

PR #374, opened 2020-09-13
Proper symbolic bound tightening for Signs

guykatzz:


src/nlr/Layer.cpp
```diff
@@ -823,6 +827,180 @@ void Layer::computeSymbolicBoundsForRelu()
...
+            _symbolicUbOfUb[i] = _eliminatedNeurons[i];
+
+            continue;
+        }
+
+        /*
+          There are two ways we can determine that a Sign has become fixed:
+
+          1. If the Sign's variable has been externally fixed
+          2. lbLb >= 0 (Positive) or ubUb <= 0 (Negative)
```
guyam2, line 855: @guykatzz if up<=0 then there is no fix because the function is non-continuous

guykatzz: Not sure I understand. Fixed here means phase fixed - if ubOfUb is negative (should be <, not <=, in the comment), then the upper bound is negative, and the Sign is in the negative phase.

guyam2 (approved):

PR #383, opened 2020-09-29
Polarity in sign constraint

anwu1219:
Currently for BNNs, only input splitting is supported in SnC mode. This PR extends the notion of polarity to Sign constraint, thus enabling the same branching heuristic as relu-splitting used in SnC for BNNs.

### DIFF
--- a/src/engine/SignConstraint.h
+++ b/src/engine/SignConstraint.h
@@ -157,7 +157,7 @@ public:
     the bound of the input to this ReLU is with respect to 0.
     Let LB be the lowerbound, and UB be the upperbound.
     If LB >= 0, polarity is 1.
-    If UB <= 0, polarity is -1.
+    If UB < 0, polarity is -1.
     If LB < 0, and UB > 0, polarity is ( LB + UB ) / (UB - LB).
 
     We divide the sum by the width of the interval so that the polarity is
--- a/src/engine/SignConstraint.h
+++ b/src/engine/SignConstraint.h
@@ -150,9 +150,40 @@ public:
     unsigned getB() const;
     unsigned getF() const;
 
+  bool supportPolarity() const;
+
+  /*
+    Return the polarity of the Sign Constraint, which computes how symmetric
+    the bound of the input to this ReLU is with respect to 0.
+    Let LB be the lowerbound, and UB be the upperbound.
+    If LB >= 0, polarity is 1.
+    If UB <= 0, polarity is -1.
+    If LB < 0, and UB > 0, polarity is ( LB + UB ) / (UB - LB).
+
+    We divide the sum by the width of the interval so that the polarity is
+    always between -1 and 1. The closer it is to 0, the more symmetric the
+    bound is.
+  */
+  double computePolarity() const;
+
+  /*
+    Update the preferred direction for fixing and handling case split
+  */
+  void updateDirection();
+
+  PhaseStatus getDirection() const;
+
+  void updateScoreBasedOnPolarity();
+
 private:
     unsigned _b, _f;
     PhaseStatus _phaseStatus;
+
+    /*
+      Denotes which case split to handle first.
+      And which phase status to repair a relu into.
+    */
+    PhaseStatus _direction;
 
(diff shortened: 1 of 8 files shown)


guykatzz:
The changes look okay to me, but we should test them on some real networks before merging.

BTW, Andrew, Guy A and I noticed the following very strange behavior. Before this PR, the default SNC heuristics on our networks was polarity-based (because of the large number of inputs). However, because none of the constraints in the network supported polarity, it would perform no splits at all. So what would happen is that the original query would run for the initial timeout; then it would run again, with a slightly longer timeout; etc. So, no splitting was occurring, and the timeouts were actually slowing us down.

And also, another question: will polarity-based work in a network where only some constraints support polarity? E.g., if we mix Maxes and Signs, where Maxes don't yet support polarity?

anwu1219:
I’m happy to run some experiments to get an idea of whether this splitting strategy works.
The behavior you observed earlier is expected due to the way splitting is implemented. But I guess a better behavior is probably using some fall back splitting strategy.
I believe the polarity-based strategy should still pick a sign constraint to split on even if there are constraints that do not support polarity.

在 2020年9月29日，下午10:16，Guy Katz <notifications@github.com> 写道：

﻿

The changes look okay to me, but we should test them on some real networks before merging.

BTW, Andrew, Guy A and I noticed the following very strange behavior. Before this PR, the default SNC heuristics on our networks was polarity-based (because of the large number of inputs). However, because none of the constraints in the network supported polarity, it would perform no splits at all. So what would happen is that the original query would run for the initial timeout; then it would run again, with a slightly longer timeout; etc. So, no splitting was occurring, and the timeouts were actually slowing us down.

And also, another question: will polarity-based work in a network where only some constraints support polarity? E.g., if we mix Maxes and Signs, where Maxes don't yet support polarity?

src/engine/SignConstraint.h
```diff
@@ -150,10 +150,41 @@ class SignConstraint : public PiecewiseLinearConstraint
...
     unsigned getF() const;
 
+  bool supportPolarity() const;
+
+  /*
+    Return the polarity of the Sign Constraint, which computes how symmetric
+    the bound of the input to this ReLU is with respect to 0.
+    Let LB be the lowerbound, and UB be the upperbound.
+    If LB >= 0, polarity is 1.
+    If UB <= 0, polarity is -1.
```
guyam2, line 160: @anwu1219 the Sign in Marabou goes by the standard that sing(x)=1 iff x>=0 so I think it should be
    **If UB < 0, polarity is -1.**

src/engine/tests/Test_SignConstraint.h
```diff
@@ -931,4 +931,63 @@ class SignConstraintTestSuite : public CxxTest::TestSuite
...
+        }
+        // b in [-2, 0], polarity should be -1, and direction should be PHASE_NEGATIVE
+        {
+            SignConstraint sign( b, f );
+            sign.notifyLowerBound( b, -2 );
+            sign.notifyUpperBound( b, 0 );
+            TS_ASSERT( sign.computePolarity() == -1 );
+
+            sign.updateDirection();
+            TS_ASSERT( sign.getDirection() == SignConstraint::PHASE_NEGATIVE );
```
guyam2, line 958: @anwu1219 @guykatzz I think this is incorrect because of the upper bound 0. Only if strictly smaller than 0 then the phase is NEGATIVE. Agree?

guyam2 (changes requested):
Only issue I have is that when the upper bound is 0 then the phase is not fixed. What do you think? And thx Andrew!

guykatzz:
Yes, that's a good point. Instead of using 0 we should probably use the \epsilon that we were using in other places in SignConstraint

src/engine/tests/Test_SignConstraint.h
anwu1219: That makes sense. I pushed a fix.

guykatzz (approved):

guyam2 (approved):

PR #391, opened 2020-10-15
Expose disjunction constraint to Python API

anwu1219:
Allow users to add disjunction constraints through the Python API

### DIFF
--- a/maraboupy/MarabouCore.cpp
+++ b/maraboupy/MarabouCore.cpp
@@ -25,7 +25,9 @@
 #include <sys/types.h>
 #include <fcntl.h>
 #include "AcasParser.h"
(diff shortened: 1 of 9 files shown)


maraboupy/MarabouCore.cpp
```diff
@@ -136,6 +138,53 @@ bool createInputQuery(InputQuery &inputQuery, std::string networkFilePath, std::
   return true;
 }
 
+std::list<unsigned> getParticipatingVariables( const Equation &equation )
```
guykatzz, line 141: So, I'm a little confused as to why this is needed. I saw you were just following what's currently going on, but do you know why this is even exposed?

guykatzz (approved):

PR #392, opened 2020-10-15
Add option to use Gurobi to solve an input query

anwu1219:
add a new command line argument --milp that calls gurobi to solve the input query.
move milp solver bound tightening type to Options.


src/common/GurobiWrapper.cpp
```diff
@@ -305,6 +310,85 @@ void GurobiWrapper::dumpModel( String name )
...
+    for ( const auto&plConstraint : inputQuery.getPiecewiseLinearConstraints() )
+    {
+        if ( !plConstraint->isActive() || plConstraint->phaseFixed() )
+            continue;
+        if ( plConstraint->getType() != PiecewiseLinearFunctionType::RELU )
+            throw MarabouError( MarabouError::UNSUPPORTED_PIECEWISE_CONSTRAINT,
+                                "GurobiWrapper::encodeInputQuery: "
+                                "Only ReLU is supported\n" );
+
+        auto relu = ( ReluConstraint *) plConstraint;
```
guykatzz, line 367: = ( ReluConstraint *) plConstraint; 
-->
= (ReluConstraint *)plConstraint;

src/common/GurobiWrapper.cpp
```diff
@@ -305,6 +310,85 @@ void GurobiWrapper::dumpModel( String name )
...
+            addGeqConstraint( terms, scalar );
+            break;
+        default:
+            break;
+        }
+    }
+
+    // Add Piecewise-linear Constraints
+    unsigned ind = 0;
+    for ( const auto&plConstraint : inputQuery.getPiecewiseLinearConstraints() )
```
guykatzz, line 358: auto& --> auto &

src/common/GurobiWrapper.cpp
```diff
@@ -305,6 +310,85 @@ void GurobiWrapper::dumpModel( String name )
...
+                     0,
+                     1,
+                     GurobiWrapper::BINARY );
+
+        unsigned sourceVariable = relu->getB();
+        unsigned targetVariable = relu->getF();
+        double sourceLb = inputQuery.getLowerBound( sourceVariable );
+        double sourceUb = inputQuery.getUpperBound( sourceVariable );
+
+        List<GurobiWrapper::Term> terms;
```
guykatzz, line 378: Can you add a comment to explain this encoding?

src/common/GurobiWrapper.h
```diff
@@ -175,6 +189,9 @@ class GurobiWrapper
     void setTimeLimit( double ) {};
     double getObjectiveBound() { return 0; };
     void dump() {}
+    void encodeInputQuery( const InputQuery & ) {}
+    String getVariableNameFromVariable( unsigned ) { return ""; };
+
```
guykatzz, line 194: remove empty line

src/engine/Engine.h
```diff
@@ -331,11 +332,22 @@ class Engine : public IEngine, public SignalHandler::Signalable
...
       Disjunction that is used for splitting but doesn't exist in the beginning
     */
     std::unique_ptr<PiecewiseLinearConstraint> _disjunctionForSplitting;
 
+    /*
+      GurobiWrapper object
+    */
+    std::unique_ptr<GurobiWrapper> _gurobi;
+
+
```
guykatzz, line 350: remove empty line

guykatzz (commented):
Apart from the comments I made inline, my main concern is that GurobiWrapper now knows about input queries. In general, classes under common/ shouldn't include things from other folders.

Also, _preprocessedQuery is just the preprocessed query - I don't like the later updating of its bounds.

So, I think the better options are:
1. Move the translation code to the IPQ. This is conceptually similar to how it constructs an NLR. Or,
2. Have this logic in the Engine, or even in a separate class, IPQtoGurobi

src/engine/MILPEncoder.cpp
```diff
@@ -0,0 +1,112 @@
...
+        if ( !plConstraint->isActive() || plConstraint->phaseFixed() )
+            continue;
+        switch ( plConstraint->getType() )
+        {
+        case PiecewiseLinearFunctionType::RELU:
+            encodeReLUConstraint( gurobi, (ReluConstraint *)plConstraint,
+                                  ind++ );
+            break;
+        default:
+            throw MarabouError( MarabouError::UNSUPPORTED_PIECEWISE_CONSTRAINT,
```
guykatzz, line 75: UNSUPPORTED_PIECEWISE_CONSTRAINT
-->
NSUPPORTED_PIECEWISE_LINEAR_CONSTRAINT

src/engine/MILPEncoder.cpp
```diff
@@ -0,0 +1,112 @@
...
+            break;
+        }
+    }
+
+    // Add Piecewise-linear Constraints
+    unsigned ind = 0;
+    for ( const auto &plConstraint : inputQuery.getPiecewiseLinearConstraints() )
+    {
+        if ( !plConstraint->isActive() || plConstraint->phaseFixed() )
+            continue;
```
guykatzz, line 67: This is a little strange. 
For phaseFixed constraint - we should probably encode them also, no? As either the identity or zero function.
for !isActive(), I'm not sure what this means in this context. Assuming MILP encoding is only invoked when Marabou is not really running, this can only happen when the preprocessor disables some constraints?

src/engine/MILPEncoder.cpp
anwu1219: fixed

src/engine/MILPEncoder.cpp
anwu1219: Before encoding the input query, we call applyValidConstraintCaseSplits(), which will update the bounds of the auxiliary variables. These bounds are encoded into gurobi earlier. So this would not be a problem in the case of a ReLU. 
But I agree the way it is now is not general. I've updated that part of the code. Please check again.

guykatzz: This feels like setting up a trap for our future selves. Lets at least add an assertion: for each relu that is either inactive or fixed, it must have an auxiliary variable, and that variable's bounds must indicate that the phase has been fixed.

anwu1219: added the assertions

guykatzz (approved):

PR #397, opened 2020-10-27
Fix a bug in Abs Constraint's bound tightening

guykatzz:


anwu1219 (approved):

vbCrLf:
Seems like it fixes the problem 👍

PR #399, opened 2020-11-02
Add context-dependent data structures from CVC4

AleksandarZeljic:
First in a series of PRs to intriduce context-dependent search implementation in Marabou. 

1. Add context-dependent (CD) data structures from CVC4 to deps folder in Marabou project. Adds all the CD data structures from CVC4, of immediate interest are Context, CDList and CDO. 

2. Fix compilation of added source files on Windows ( variadic syntax, /deps/base/check.h used some g++ specific pre-processor instructions). Noteworthy for the future, if we ever update CD files with new versions from CVC4 (due to bug fixes, updates, etc).


deps/CVC4/include/cvc4_private.h
```diff
@@ -0,0 +1,55 @@
+/*********************                                                        */
+/*! \file cvc4_public.h
```
ahmed-irfan, line 2: IIRC, we modified this file. It was a hack. 
Now the file header and its header guard do not match with the file name. 
Maybe, we should try to have a clean solution for this.

deps/CVC4/include/cvc4_private_library.h
```diff
@@ -0,0 +1,55 @@
+/*********************                                                        */
+/*! \file cvc4_public.h
```
ahmed-irfan, line 2: ditto

CMakeLists.txt
```diff
@@ -28,6 +28,8 @@ set(MARABOU_TEST_LIB MarabouHelperTest)
 set(MARABOU_EXE Marabou${CMAKE_EXECUTABLE_SUFFIX})
 set(MARABOU_PY MarabouCore)
 
+
```
anwu1219, line 31: Extra spaces and empty lines in this file.

deps/CVC4/include/cvc4_private.h
AleksandarZeljic: Good catch, will fix!

CMakeLists.txt
AleksandarZeljic: Thanks, will make spacing uniform throughout the file.

ahmed-irfan (approved):
LGTM.

PR #400, opened 2020-11-02
Fix in property parser

anwu1219:
The check for scalar on the right hand side throws error on scientific notations. This PR fixes that.


src/input_parsers/PropertyParser.cpp
```diff
@@ -20,15 +20,15 @@
 #include "PropertyParser.h"
 #include <regex>
```
ahmed-irfan, line 21: check if regex is needed now?

anwu1219: Removed now

ahmed-irfan (approved):
LGTM. 

If we can do something to improve the code-coverage, that would be great.

PR #402, opened 2020-11-05
Add Guroby encoder for max constratints.

tagomaru:

This PR is going to add Guroby encoder for max constraints.

## Sample NNET:
```
// simple_model 
2,2,3,3,
2,2,3,
0,
-100000.0,-100000.0,
100000.0,100000.0,
0.0,0.0,0.0,
1.0,1.0,1.0,
1.0,1.0,
-1.0,-1.0,
0.0,
0.0,
2.0,-1.0,
-1.0,1.0,
1.0,-1.0,
0.0,
0.0,
0.0,
```

## Sample Code:
```
from maraboupy import Marabou as marabou
nnet_file_name = "./simple_model.nnet"
net = marabou.read_nnet(nnet_file_name)
net.setLowerBound(net.outputVars[0][0], 2)
net.setUpperBound(net.outputVars[0][0], 2)
net.setLowerBound(net.inputVars[0][0], 1)
net.setUpperBound(net.inputVars[0][0], 1)
net.addMaxConstraint(set(net.outputVars[0]), net.outputVars[0][0])
options = marabou.createOptions(solveWithMILP=True)
vals, stats = net.solve(options=options)
```
## Test with the above sample code:
command
```
python3 guroby_max.py
```

result
```
...

sat
input 0 = 1.0
input 1 = -0.0
output 0 = 2.0
output 1 = -1.0
output 2 = 1.0
```

max


src/engine/MILPEncoder.cpp
```diff
@@ -126,3 +129,85 @@ void MILPEncoder::encodeReLUConstraint( GurobiWrapper &gurobi, ReluConstraint
...
+    // upper bounds of each x_i
+    double* ubs = new double[m];
+
+    // terms for Gurobi
+    List<GurobiWrapper::Term> terms;
+
+    unsigned i = 0;
+    for ( const auto &x : xs ) {
+        // add binary variable
+        gurobi.addVariable( Stringf( "a%u", x ),
```
anwu1219, line 155: Naming the auxiliary variable like this would result in duplicated variable names. For instance, if we have two Max constraints Max([x1,x2], x4) and Max([x1,x3], x5), we would introduce the variable a1 twice, while we actually need two fresh variables. 
One solution is to keep track of the latest index of the auxiliary variable. Please check the encodeReLUConstraint() method above.

src/engine/MILPEncoder.cpp
```diff
@@ -126,3 +129,85 @@ void MILPEncoder::encodeReLUConstraint( GurobiWrapper &gurobi, ReluConstraint
...
+        double umax = getUmax( ubs, i, m );
+        terms.append( GurobiWrapper::Term( 1, Stringf( "x%u", y ) ) );
+        terms.append( GurobiWrapper::Term( -1, Stringf( "x%u", x ) ) );
+        terms.append( GurobiWrapper::Term( umax - _tableau.getLowerBound( x ), Stringf( "a%u", x ) ) );
+        gurobi.addLeqConstraint( terms, umax - _tableau.getLowerBound( x ) );
+
+        terms.clear();
+
+        // add constraint: y >= x_i
+        terms.append( GurobiWrapper::Term( 1, Stringf( "x%u", y ) ) );
```
anwu1219, line 182: "y >= x_i" is already encoded as an inequality here: https://github.com/NeuralNetworkVerification/Marabou/blob/3bf23beb8a9ce3d2724bf63984af792ec006c370/src/engine/MaxConstraint.cpp#L489-L515.
I think you could just remove line 182 - 188.

src/engine/MILPEncoder.cpp
```diff
@@ -126,3 +129,85 @@ void MILPEncoder::encodeReLUConstraint( GurobiWrapper &gurobi, ReluConstraint
...
+            continue;
+        }
+
+        if ( ubs[umaxIndex] < ubs[i] ) {
+            umaxIndex = i;
+        }
+    }
+
+    return ubs[umaxIndex];  
+}
```
anwu1219, line 213: Add a newline here

src/engine/MILPEncoder.cpp
```diff
@@ -126,3 +129,85 @@ void MILPEncoder::encodeReLUConstraint( GurobiWrapper &gurobi, ReluConstraint
...
+
+    // add constraint: a_1 + a_2 + ... + a_m = 1
+    gurobi.addEqConstraint( terms, 1 );
+
+    terms.clear();
+
+    i = 0;
+    for ( const auto &x : xs ) {
+        // add constraint: y <= x_i + (1 - a_i) * (umax - l)
+        double umax = getUmax( ubs, i, m );
```
anwu1219, line 173: if we store the <upper bound, variable> pair in a priority queue, we should be able to compute each Umax in O(log n) time. Currently it is computed in linear time, which might create overhead as the network size increases.

anwu1219 (changes requested):

src/engine/MILPEncoder.cpp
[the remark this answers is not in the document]
tagomaru: will do.

src/engine/MILPEncoder.cpp
tagomaru: will do.

src/engine/MILPEncoder.cpp
tagomaru: will do.

src/engine/MILPEncoder.cpp
tagomaru: will do.

src/engine/MILPEncoder.cpp
tagomaru: Your idea may work, but I do not think it makes sense to me at the moment.
I think your idea needs to generate priority queue per [this loop](https://github.com/NeuralNetworkVerification/Marabou/blob/35fd279a3d018dd1b1791b630fc762d2e08c7299/src/engine/MILPEncoder.cpp#L171-L189) before calling _getUmax_ since the priorities are different per _i_.
However it requires log(n) to push elements into the priority queue for the each loop anyway.

I tried to decrease the computation cost by sharing upper bound array for every _i_.

If my understanding is not correct, please let me know.

What do you think?

tagomaru: I did.

src/engine/MILPEncoder.cpp
tagomaru: done.

src/engine/MILPEncoder.cpp
tagomaru: done.

src/engine/MILPEncoder.cpp
tagomaru: done.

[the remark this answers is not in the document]
tagomaru: done.

tagomaru:
I changed this status to WIP since Test_preprocessor failed after merging master.
```
12/58 Test #16: Test_Preprocessor .................***Failed    1.07 sec
Assertion violation! File /Users/teruhiro/Documents/work/Marabou/src/engine/InputQuery.cpp, line 279
```

tagomaru:
# Evaluation result:
## 1. MILP
### Model
The onnx file (`resources/onnx/cnn_max_mninst2.onnx`) of the model was converted from `resources/keras/cnn_max_mnist2.h5`.
```
model = models.Sequential()
model.add(layers.Conv2D(8, (4, 4), activation='relu', input_shape=(28, 28, 1)))
model.add(layers.Conv2D(4, (4, 4), activation='relu'))
model.add(layers.MaxPooling2D((2, 2)))
model.add(layers.Flatten())
model.add(layers.Dense(32, activation='relu'))
model.add(layers.Dense(10, activation='softmax'))
```

### Test program (onnx_cnn_max_mnist2.py)
``` python
from maraboupy import Marabou
import numpy as np

options = Marabou.createOptions(verbosity = 2, solveWithMILP=True)
filename = 'cnn_max_mninst2.onnx'
network = Marabou.read_onnx(filename)

inputVars = network.inputVars[0][0]
outputVars = network.outputVars[0]

delta = 0.1
for h in range(inputVars.shape[0]):
    for w in range(inputVars.shape[1]):
        network.setLowerBound(inputVars[h][w][0], 0.5-delta)
        network.setUpperBound(inputVars[h][w][0], 0.5+delta)

# network.setLowerBound(outputVars[0], 1.0)

vals, stats = network.solve(options = options)
assert len(vals) > 0

dataONNX = np.zeros(inputVars.shape)
for i in range(inputVars.shape[0]):
    for j in range(inputVars.shape[1]):
        dataONNX[i][j][0] = vals[28 * i + j]

onnxEval = network.evaluateWithoutMarabou([dataONNX])

print('\nonnx output:\n', onnxEval)
```

### Execution
```bash
$ /usr/bin/time -l python3 onnx_cnn_max_mnist2.py
(removed)
output 0 = -0.11791291526924269
output 1 = -0.1608994968618801
output 2 = -3.053131166312996
output 3 = -1.9990824676561814
output 4 = 1.0521385098435299
output 5 = -1.4255639982065038
output 6 = 0.02008915031516549
output 7 = -1.215866175485384
output 8 = -0.7083881322782134
output 9 = -0.47855364129358235

onnx output:
 [[-0.11791354 -0.1609008  -3.0531301  -1.9990811   1.0521375  -1.4255633
   0.02008897 -1.2158666  -0.7083874  -0.47855455]]
    96692.76 real     93300.76 user       412.88 sys
12301275136  maximum resident set size
         0  average shared memory size
         0  average unshared data size
         0  average unshared stack size
   8717705  page reclaims
      1576  page faults
         0  swaps
         0  block input operations
         0  block output operations
         0  messages sent
         0  messages received
         0  signals received
     12856  voluntary context switches
  19258691  involuntary context switches
```
### Elapsed time
Total 27 hours. (20 hours until preprocess was done)

src/engine/InputQuery.cpp
```diff
@@ -255,18 +255,28 @@ InputQuery &InputQuery::operator=( const InputQuery &other )
...
-                 auto *newPlc = constraint->duplicateConstraint();
-                 _plConstraints.append( newPlc );
-                 ++numberOfDisjunctions;
-             }
+            {
+                auto *newPlc = constraint->duplicateConstraint();
+                _plConstraints.append( newPlc );
+                ++numberOfDisjunctions;
+            }
+            else if ( constraint->getType() == MAX &&
```
anwu1219, line 267: I think we assume that network level reasoner can pick up all the Max constraint. Maybe just remove the logic for Max here?

src/engine/MILPEncoder.cpp
```diff
@@ -118,11 +118,70 @@ void MILPEncoder::encodeReLUConstraint( GurobiWrapper &gurobi, ReluConstraint
...
+    auto cmp = []( qtype l, qtype r) { return l.first <= r.first; };
+    std::priority_queue<qtype, std::vector<qtype>, decltype( cmp )> ubq( cmp );
+
+    // terms for Gurobi
+    List<GurobiWrapper::Term> terms;
+
+    for ( const auto &x : xs ) 
+    {
+        // add binary variable
+        gurobi.addVariable( Stringf( "a%u_%u", _binVarIndex, x ),
```
anwu1219, line 152: Please explain the name convention for future reference.

src/engine/MILPEncoder.h
```diff
@@ -51,6 +51,8 @@ class MILPEncoder
     */
     Map<unsigned, String> _variableToVariableName;
 
+    unsigned _binVarIndex = 0;
```
anwu1219, line 54: Add documentation.

anwu1219 (approved):

src/engine/InputQuery.cpp
tagomaru: Your understanding is right.
However, when the network reasoner is true, the below part also picks up max constraints.
To avoid to pick up doubly, this neglects max constraints here.

```python
        for ( const auto &constraint : other._networkLevelReasoner->
                  getConstraintsInTopologicalOrder() )
        {
            auto *newPlc = constraint->duplicateConstraint();
            _plConstraints.append( newPlc );
            _networkLevelReasoner->addConstraintInTopologicalOrder( newPlc );
        }
```

Does it make sense?

src/engine/MILPEncoder.cpp
tagomaru: I did.

src/engine/MILPEncoder.h
tagomaru: I did.

PR #403, opened 2020-11-09
Fix a compilation issue on mac OS

alexopenu:
This pr addresses a similar compilation error on a mac to the one that was fixed in #294. 

Like before, including cmath solves the ambiguity issue of the call to abs(). @guykatzz, do you think it would make more sense to include it in a different header, more globally (e.g., FloatUtils)?

### DIFF
--- a/src/common/FloatUtils.cpp
+++ b/src/common/FloatUtils.cpp
@@ -54,12 +54,12 @@ String FloatUtils::doubleToString( double x, unsigned precision )
 
 bool FloatUtils::isNan( double x )
 {
-    return isnan( x );
+    return std::isnan( x );
 }
 
 bool FloatUtils::isInf( double x )
 {
-    return isinf( x );
+    return std::isinf( x );
 }
 
 //
--- a/src/common/FloatUtils.h
+++ b/src/common/FloatUtils.h
@@ -21,6 +21,7 @@
 #include "MString.h"
 
 #include <cfloat>
+#include <cmath>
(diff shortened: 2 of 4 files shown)


guykatzz:
Hi Alex,
Thanks for the PR. Yes, we can just put it in FloatUtils.

guykatzz (approved):

alexopenu:
> Hi Alex,
> Thanks for the PR. Yes, we can just put it in FloatUtils.

Hi Guy, 

Now that I moved <cmath> to FloatUtils, travis build fails... However, it builds perfectly well both on my OS and on the linux distribution in my office. Any ideas? We can, of course, always move the include back to SignConstraint...

guykatzz:
Hi Alex,
I think we should debug this.

Are you using the exact same flags as Travis? Did you try with both the gcc
and clang compilers?
Did you Google the errors?
From a quick look, you're including both <cmath> and <math.h>; I think the
former is the "c++ variant"  of the latter, and I'm not sure whether
including them both can cause strange behaviors.

alexopenu:
Sure, I just wasn't sure whether you preferred making a bigger change, given that it had worked fine with math.h except for one problem that could be solved locally; but I agree that it is better to solve the issue globally. I think it is fine now, all tests pass, so if you are happy, I will merge it.

(sorry, accidentally closed and reopened the pr earlier)

guykatzz:
Looks good, you can go ahead and merge.

PR #404, opened 2020-11-12
Refactor _phaseStatus from individual classes into PiecewiseLinearConstraint

AleksandarZeljic:
This PR:
* introduces a joint PhaseStatus enumeration for all PiecewiseLinearConstraints
* refactors _phaseStatus member from derived classes into PiecewiseLinearConstraint
* refactors usage of local PhaseStatus enumerations with the global enumeration
* removes local PhaseStatus enumerations

To add new classes inheriting PiecewiseLinearConstraint their phase status elements need to be added to the PhaseStatus enumeration. Classes like Disjunction or MaxConstraint use PhaseStatus enumeration as an unsigned value instead of the label.

### DIFF
--- a/src/engine/PiecewiseLinearConstraint.h
+++ b/src/engine/PiecewiseLinearConstraint.h
@@ -39,7 +39,7 @@ enum PhaseStatus : unsigned {
     ABS_PHASE_NEGATIVE = 4,
     SIGN_PHASE_POSITIVE = 5,
     SIGN_PHASE_NEGATIVE = 6
-;
+};
 
 class PiecewiseLinearConstraint : public ITableau::VariableWatcher
 {
--- a/src/engine/PiecewiseLinearConstraint.h
+++ b/src/engine/PiecewiseLinearConstraint.h
@@ -265,7 +265,8 @@ public:
 
 protected:
     bool _constraintActive;
-	Map<unsigned, double> _assignment;
+    PhaseStatus _phaseStatus;
+    Map<unsigned, double> _assignment;
     Map<unsigned, double> _lowerBounds;
     Map<unsigned, double> _upperBounds;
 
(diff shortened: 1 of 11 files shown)


src/engine/PiecewiseLinearConstraint.h
```diff
@@ -31,6 +31,17 @@ class ITableau;
...
 
+enum PhaseStatus : unsigned {
+    PHASE_NOT_FIXED = 0,
+    RELU_PHASE_ACTIVE = 1,
+    RELU_PHASE_INACTIVE = 2,
+    ABS_PHASE_POSITIVE = 3,
+    ABS_PHASE_NEGATIVE = 4,
+    SIGN_PHASE_POSITIVE = 5,
+    SIGN_PHASE_NEGATIVE = 6
+
```
guykatzz, line 42: Remove empty line

guykatzz (commented):
What's the motivation for this change? In general, it is not desirable for a parent class to store information about the child classes.

src/engine/PiecewiseLinearConstraint.h
AleksandarZeljic: All PiecewiseLinearConstraints have a phase/case, either a fixed number of them (ReLU, Sign, Abs) or a flexible number of them (Max, Disjunction). The first  group cares about named elements of the enumeration, while the second group cares about their numerical value, and a fixed size enumeration caters to both. Regardless of the subclass, the role of these fields is the same and becomes further relevant in context-dependent implementation for detection of implications, contradictions, etc.   

The idea is to generalize this field to avoid repetition of identical code in child classes, and level the implementation so that a single intermediate child class implement the context-dependent behavior, while retaining existing functionality for preprocessing purposes. (The intermediate class can be merged with the parent class at a later date). 

Implementation-wise, we can make PiecewiseLinearConstraint a template class and provide the individual PhaseStatus enumerations as the parameter. However, instead of one global enumeration, there would be several smaller enumerations, which would be outside of their child template classes. This solution has some small upsides for type-checking, but adds some overhead for implementation. 

For general overview, the next PR adds ContextDependentPiecewiseLinearConstraint, which inherits PiecewiseLinearConstraint and  introduces context-dependent and lazily backtracking members _cdConstraintActive, _cdPhaseStatus and uniformly implement relevant methods which use parent class members until the context-based members are initialized. Current architecture would require that child classes implement individually context-dependent behavior, which would be undesirable given the delicacy of context-dependent data structures and causing additional code repetition. Finally, this approach allows for simple transition to context-based architecture (when the time comes), by inheriting the ContextDependentPiecewiseLinearConstraint class and transitively the PiecewiseLinearConstraint class. If needed I can include that PR with this.

guykatzz (approved):

guykatzz:
Okay :)

PR #408, opened 2020-11-26
support user defined preprocessor bound tolerance in marabou/py api

yizhake:
Currently, Preprocessor uses a default value of epsilon from GlobalConfiguration for bound tightening.
The code of this PR enables to define this epsilon's value from command line and while using maraboupy API.

### DIFF
--- a/maraboupy/Marabou.py
+++ b/maraboupy/Marabou.py
@@ -120,7 +120,8 @@ def solve_query(ipq, filename="", verbose=True, options=None):
 def createOptions(numWorkers=1, initialTimeout=5, initialDivides=0, onlineDivides=2,
                   timeoutInSeconds=0, timeoutFactor=1.5, verbosity=2, snc=False,
                   splittingStrategy="auto", sncSplittingStrategy="auto",
-                  restoreTreeStates=False, splitThreshold=20, solveWithMILP=False ):
+                  restoreTreeStates=False, splitThreshold=20, solveWithMILP=False,
+                  preprocessorBoundTolerance=0.0000000001):
     """Create an options object for how Marabou should solve the query
 
     Args:
@@ -138,6 +139,7 @@ def createOptions(numWorkers=1, initialTimeout=5, initialDivides=0, onlineDivide
         sncSplittingStrategy (string, optional): Specifies which partitioning strategy to use in the DNC mode (auto/largest-interval/polarity).
         restoreTreeStates (bool, optional): Whether to restore tree states in dnc mode, defaults to False
         solveWithMILP ( bool, optional): Whther to solve the input query with a MILP encoding. Currently only works when Gurobi is installed. Defaults to False.
+        preprocessorBoundTolerance ( float, optional): epsilon value for preprocess bound tightening . Defaults to 10^-10.
     Returns:
         :class:`~maraboupy.MarabouCore.Options`
     """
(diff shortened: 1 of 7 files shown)


src/engine/Engine.cpp
```diff
@@ -266,6 +266,7 @@ bool Engine::solve( unsigned timeoutInSeconds )
                 continue;
             }
 
+
```
guykatzz, line 269: Remove new empty line

src/engine/Preprocessor.cpp
```diff
@@ -336,7 +339,8 @@ bool Preprocessor::processEquations()
 
                 lowerBound /= -ci;
 
-                if ( FloatUtils::gt( lowerBound, _preprocessed.getLowerBound( xi ) ) )
+                double diff = lowerBound - _preprocessed.getLowerBound( xi );
+                if ( FloatUtils::isPositive( diff, epsilon ) )
```
guykatzz, line 343: Switch back to gt( x, y, epsilon ) to make it more readable

src/engine/Preprocessor.cpp
```diff
@@ -360,7 +364,8 @@ bool Preprocessor::processEquations()
 
                 upperBound /= -ci;
 
-                if ( FloatUtils::lt( upperBound, _preprocessed.getUpperBound( xi ) ) )
+                double diff = upperBound - _preprocessed.getUpperBound( xi );
+                if ( FloatUtils::isNegative( diff, epsilon ) )
```
guykatzz, line 368: same as above

maraboupy/Marabou.py
```diff
@@ -120,7 +120,8 @@ def solve_query(ipq, filename="", verbose=True, options=None):
 def createOptions(numWorkers=1, initialTimeout=5, initialDivides=0, onlineDivides=2,
                   timeoutInSeconds=0, timeoutFactor=1.5, verbosity=2, snc=False,
                   splittingStrategy="auto", sncSplittingStrategy="auto",
-                  restoreTreeStates=False, splitThreshold=20, solveWithMILP=False ):
+                  restoreTreeStates=False, splitThreshold=20, solveWithMILP=False,
+                  preprocessorBoundTolerance=0.0000000001):
```
guykatzz, line 124: This is the same initialization value as in Options.cpp, right? Why do we need to initialize in two places? If we leave it uninitialized here, won't it default to the value in Options.cpp?

src/engine/Engine.cpp
yizhake: Done

src/engine/Preprocessor.cpp
yizhake: Done

src/engine/Preprocessor.cpp
yizhake: Done

maraboupy/Marabou.py
yizhake: This value must be initialized with any value, since it is the (not the first and actually the) last parameter in a list of initialized function's parameters. Otherwise, the interpreter raise an exception.

In addition, if I move it to be the first parameter in the function it can make damage to every call without this parameter. It will not backward compatible, and also won't use the default value from the cpp file. It is possible, but in my opinion, the current option is better.

Maybe it is possible to take the default value using some python-cpp API, but it seems that default values can't be supported (at least while using pybind, since they are not part of the signature of the function in cpp). Anyway, the whole function uses default variables which are defined in the cpp file too, so i think it should be solved in another PR.

guykatzz (approved):

PR #420, opened 2021-01-06
Deeppoly analysis

anwu1219:
Adding support for deeppoly analysis for ReLU activation described in https://files.sri.inf.ethz.ch/website/papers/DeepPoly.pdf.

1. Currently supporting six abstract transformers (DeepPoly{ReLU, WeightedSum, Input, MaxPool, Absolute, Sign}Element.h). 
2. DeepPolyAnalysis.h contains the method to run the abstract interpretation.
3. Adding an option to choose whether to use sbt or deeppoly or none during the search.
4. Adding an option to print out bounds for each neuron after preprocessing (this would help comparing different bound tightening passes).


maraboupy/MarabouCore.cpp
```diff
@@ -218,11 +221,14 @@ struct MarabouOptions {
     // string options
     Options::get()->setString( Options::SPLITTING_STRATEGY, _splittingStrategyString );
     Options::get()->setString( Options::SNC_SPLITTING_STRATEGY, _sncSplittingStrategyString );
+    Options::get()->setString( Options::SYMBOLIC_BOUND_TIGHTENING_TYPE, _tighteningStrategyString );
+
```
guykatzz, line 225: remove empty line

src/nlr/DeepPolySignElement.cpp
```diff
@@ -0,0 +1,231 @@
...
+            // Concrete bound: 1 <= x_f <= 1
+            _symbolicUb[i] = 0;
+            _symbolicUpperBias[i] = 1;
+            _ub[i] = 1;
+
+            _symbolicLb[i] = 0;
+            _symbolicLowerBias[i] = 1;
+            _lb[i] = 1;
+        }
+        else if ( !FloatUtils::isPositive( sourceUb ) )
```
guykatzz, line 65: Signs are not continuous. We're in the negative phase if the input is negative, and in the positive phase if it's non negative. Better change this to isNegative (instead of !isPositive)

maraboupy/MarabouCore.cpp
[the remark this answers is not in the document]
anwu1219: resolved

src/engine/Engine.cpp
```diff
@@ -1106,7 +1107,13 @@ bool Engine::processInputQuery( InputQuery &inputQuery, bool preprocess )
         delete[] constraintMatrix;
 
         if ( preprocess )
+        {
+            performSymbolicBoundTightening();
```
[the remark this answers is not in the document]
anwu1219: I'm not sure whether LP derives strictly tighter bounds. But I think it makes sense to use sbt/deepoly to fix ReLUs first to reduce the number of LP calls (we do not solve LP for ReLUs that are already fixed).

src/nlr/DeepPolyAnalysis.cpp
```diff
@@ -0,0 +1,206 @@
...
+    {
+        /*
+          Go over the layers, one by one. Each time construct and execute
+          the abstract element.
+        */
+        ASSERT( layers.exists( i ) );
+        log( Stringf( "Running deeppoly analysis for layer %u...", i ) );
+
+        Layer *layer = layers[i];
+        DeepPolyElement *deepPolyElement = createDeepPolyElement( layer );
```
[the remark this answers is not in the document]
anwu1219: Changed the code such that elements are constructed once.

src/nlr/DeepPolyMaxPoolElement.cpp
```diff
@@ -0,0 +1,227 @@
...
+            }
+        }
+
+        // The phase is fixed if the lower-bound of a source variable x_b is
+        // larger than the upper-bounds of the other source variables.
+        bool phaseFixed = true;
+        for ( const auto &sourceIndex : sources )
+        {
+            if ( sourceIndex != indexOfMaxLowerBound &&
+                 FloatUtils::gt( sourceUbs[sourceIndex], maxLowerBound ) )
```
guykatzz, line 81: shouldn't this be lt?

src/nlr/DeepPolyMaxPoolElement.cpp
```diff
@@ -0,0 +1,227 @@
...
+    freeMemoryIfNeeded();
+}
+
+void DeepPolyMaxPoolElement::execute( const Map<unsigned, DeepPolyElement *>
+                               &deepPolyElementsBefore )
+{
+    log( "Executing..." );
+    ASSERT( hasPredecessor() );
+    allocateMemory();
+
```
guykatzz, line 39: If I understood correctly, for the phase fixed case, max is just the identity function; and for the phase not fixed case, the symbolic ub is concretized, and the symbolic lb is the identity for the element with the largest lb?

We can probably use a symbolic expression for the UB as well. There's the work Matan Ostrovsky is doing, and there's also a suggestion the Ehler's Planet paper from 2017.

src/nlr/DeepPolyMaxPoolElement.cpp
anwu1219: Whoops yes. I'll add more test for max constraints.

src/nlr/DeepPolyMaxPoolElement.cpp
anwu1219: >> for the phase fixed case, max is just the identity function; and for the phase not fixed case, the symbolic ub is concretized, and the symbolic lb is the identity for the element with the largest lb?
<< Yes, that's the formulation in the original paper. 

I'll check with Matan about it. Thanks for the pointer!

src/nlr/DeepPolyMaxPoolElement.cpp
anwu1219: > Whoops yes. I'll add more test for max constraints.

Actually. I think the code is right. The max constraint is fixed if there is a input variable whose lower bound is greater than the upper bound of the other input variables. In that case, we can conclude that the Max is fixed.

guykatzz (approved):

PR #422, opened 2021-01-06
Context Dependent Piecewise Linear Constraint

AleksandarZeljic:
This PR introduces an implementation of PiecewiseLinearConstraint using
context-dependent (CD) data structures from CVC4, which lazily and automatically
backtrack in sync with the central context object. These data structures
are delicate and require that data stored is safe to memcopy around. 
 
Members are initialized using the initializeCDOs( Context context ) method.
Subclasses need to call initializeDuplicatesCDOs() method in the
duplicateConstraint() method to avoid sharing context-dependent members.
CDOs need not be initialized for pre-processing purposes.

The class uses CDOs to record:
   * _cdConstraintActive
   * _cdPhaseStatus
   * _cdInfeasibleCases
 
PhaseStatus enumeration is used to keep track of current phase of the
object, as well as to record previously explored phases. This replaces
moving PiecewiseLinearCaseSplits around.
 
ContextDependentPiecewiseLinearConstraint stores locally which phases have
been explored so far using _cdInfeasibleCases. To communicate with the
search it implements two methods:
   * markInfeasible( PhaseStatus caseId ) - marks explored phase
   * nextFeasibleCase() - obtains next unexplored case if one exists
 
These methods rely on subclass properly handling:
   * _numCases field, which denotes the number of piecewise linear
      segments the constraint has, initialized by the constructor.
   * getAllCases() virtual method returns all possible phases represented using
     PhaseStatus enumeration. The order of returned cases will determine the
     search order, so this method should implement any search heuristics.

### DIFF
--- a/src/engine/ContextDependentPiecewiseLinearConstraint.cpp
+++ b/src/engine/ContextDependentPiecewiseLinearConstraint.cpp
@@ -162,7 +162,7 @@ PhaseStatus ContextDependentPiecewiseLinearConstraint::nextFeasibleCase()
     ASSERT( getPhaseStatus() == PHASE_NOT_FIXED );
 
     if ( !isFeasible() )
-        return PHASE_NOT_FIXED;
+        return CONSTRAINT_INFEASIBLE;
 
     List<PhaseStatus> allCases = getAllCases();
     for ( PhaseStatus thisCase : allCases )
@@ -170,14 +170,14 @@ PhaseStatus ContextDependentPiecewiseLinearConstraint::nextFeasibleCase()
         auto loc =
             std::find( _cdInfeasibleCases->begin(), _cdInfeasibleCases->end(), thisCase );
 
-        // Case not found, therefore it is feasible
+        // Case is not infeasible, return it as feasible
         if ( loc == _cdInfeasibleCases->end() )
             return thisCase;
     }
 
     // UNREACHABLE
     ASSERT( false );
-    return PHASE_NOT_FIXED;
+    return CONSTRAINT_INFEASIBLE;
 }
 
 //
--- a/src/engine/ContextDependentPiecewiseLinearConstraint.cpp
+++ b/src/engine/ContextDependentPiecewiseLinearConstraint.cpp
@@ -71,26 +71,6 @@ void ContextDependentPiecewiseLinearConstraint::initializeCDOs(
     initializeCDInfeasibleCases();
 }
 
-void ContextDependentPiecewiseLinearConstraint::cdoCleanup()
-{
-    if ( nullptr != _cdConstraintActive )
-        _cdConstraintActive->deleteSelf();
-
-    _cdConstraintActive = nullptr;
-
-    if ( nullptr != _cdPhaseStatus )
-        _cdPhaseStatus->deleteSelf();
-
-    _cdPhaseStatus = nullptr;
-
-    if ( nullptr != _cdInfeasibleCases )
-        _cdInfeasibleCases->deleteSelf();
-
-    _cdInfeasibleCases = nullptr;
-
-    _context = nullptr;
-}
-
 void ContextDependentPiecewiseLinearConstraint::initializeCDInfeasibleCases()
 {
     ASSERT( nullptr != _context );
@@ -111,6 +91,26 @@ void ContextDependentPiecewiseLinearConstraint::initializeCDPhaseStatus()
     ASSERT( nullptr == _cdPhaseStatus );
     _cdPhaseStatus =
         new ( true ) CVC4::context::CDO<PhaseStatus>( _context, PHASE_NOT_FIXED );
+}
+
+void ContextDependentPiecewiseLinearConstraint::cdoCleanup()
+{
+    if ( nullptr != _cdConstraintActive )
+        _cdConstraintActive->deleteSelf();
+
+    _cdConstraintActive = nullptr;
+
+    if ( nullptr != _cdPhaseStatus )
+        _cdPhaseStatus->deleteSelf();
+
+    _cdPhaseStatus = nullptr;
+
+    if ( nullptr != _cdInfeasibleCases )
+        _cdInfeasibleCases->deleteSelf();
+
+    _cdInfeasibleCases = nullptr;
+
+    _context = nullptr;
 }
 
 PhaseStatus ContextDependentPiecewiseLinearConstraint::getPhaseStatus() const
--- a/src/engine/ContextDependentPiecewiseLinearConstraint.cpp
+++ b/src/engine/ContextDependentPiecewiseLinearConstraint.cpp
@@ -159,11 +159,10 @@ void ContextDependentPiecewiseLinearConstraint::markInfeasible(
 
 PhaseStatus ContextDependentPiecewiseLinearConstraint::nextFeasibleCase()
 {
+    ASSERT( getPhaseStatus() == PHASE_NOT_FIXED );
+
     if ( !isFeasible() )
         return PHASE_NOT_FIXED;
-
-    if ( phaseFixed() )
-        return getPhaseStatus();
 
     List<PhaseStatus> allCases = getAllCases();
     for ( PhaseStatus thisCase : allCases )
--- a/src/engine/ContextDependentPiecewiseLinearConstraint.cpp
+++ b/src/engine/ContextDependentPiecewiseLinearConstraint.cpp
@@ -82,7 +82,7 @@ void ContextDependentPiecewiseLinearConstraint::initializeCDActiveStatus()
 {
     ASSERT( _context != nullptr );
     ASSERT( _cdConstraintActive == nullptr );
-    _cdConstraintActive = new ( true ) CVC4::context::CDO<bool>( _context, true );
+    _cdConstraintActive = new ( true ) CVC4::context::CDO<bool>( _context, _constraintActive );
 }
 
 void ContextDependentPiecewiseLinearConstraint::initializeCDPhaseStatus()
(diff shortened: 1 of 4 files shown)


src/engine/ContextDependentPiecewiseLinearConstraint.h
```diff
@@ -0,0 +1,246 @@
...
+#include "context/context.h"
+
+class Equation;
+class BoundManager;
+class ITableau;
+class InputQuery;
+class String;
+
+class ContextDependentPiecewiseLinearConstraint
+   : public virtual PiecewiseLinearConstraint
```
guykatzz, line 70: is virtual inheritance needed here?

src/engine/ContextDependentPiecewiseLinearConstraint.h
```diff
@@ -0,0 +1,246 @@
...
+    virtual List<PhaseStatus> getAllCases() const = 0;
+
+    /*
+     * Returns case split corresponding to the given phase/id
+       TODO: Update the signature in PiecewiseLinearConstraint, once the new
+       search is integrated.
+     */
+    virtual PiecewiseLinearCaseSplit getCaseSplit( PhaseStatus caseId ) const = 0;
+
+
```
guykatzz, line 102: remove empty line

src/engine/ContextDependentPiecewiseLinearConstraint.h
```diff
@@ -0,0 +1,246 @@
...
+    /*
+      Check if the constraint's phase has been fixed.
+    */
+    virtual bool phaseFixed() const = 0;
+
+    /*
+      If the constraint's phase has been fixed, get the (valid) case split.
+    */
+    virtual PiecewiseLinearCaseSplit getImpliedCaseSplit() const = 0;
```
guykatzz, line 111: As a side note, we said we wanted to change this name, right? Maybe getImpliedCase, to be consistent with getAllCases()?

src/engine/ContextDependentPiecewiseLinearConstraint.h
```diff
@@ -0,0 +1,246 @@
...
+
+    /*
+       Politely clean up allocated CDOs.
+     */
+    void cdoCleanup();
+
+    /*
+      Get the context object - debugging purposes only
+    */
+    CVC4::context::Context *getContext() const
```
guykatzz, line 135: Can we make the returned object const? i.e., const Context *
Same for the other debugging functions that return pointers

src/engine/ContextDependentPiecewiseLinearConstraint.h
```diff
@@ -0,0 +1,246 @@
...
+    void markInfeasible( PhaseStatus exploredCase );
+
+    /*
+      Returns phaseStatus representing next feasible case.
+      Returns PHASE_NOT_FIXED if no feasible case exists.
+      Assumes the contraint phase is PHASE_NOT_FIXED.
+      Worst case complexity O(n^2)
+      This method should be overloaded for MAX and DISJUNCTION constraints.
+     */
+    PhaseStatus nextFeasibleCase();
```
guykatzz, line 176: If the method should be overloaded, shouldn't it be declared virtual?

src/engine/ContextDependentPiecewiseLinearConstraint.h
```diff
@@ -0,0 +1,246 @@
...
+
+    /*
+       Returns number of cases not yet marked as infeasible.
+     */
+    unsigned numFeasibleCases()
```
guykatzz, line 181: make function const?

src/engine/ContextDependentPiecewiseLinearConstraint.h
```diff
@@ -0,0 +1,246 @@
...
+     */
+    bool isFeasible()
+    {
+        return numFeasibleCases() > 0u;
+    }
+
+    /*
+       Returns true if there is only one feasible case remaining.
+     */
+    bool isImplication()
```
guykatzz, line 197: Make function const?
Also, we usually do == 1 ad not 1 ==.

src/engine/ContextDependentPiecewiseLinearConstraint.cpp
```diff
@@ -0,0 +1,189 @@
...
+    if ( nullptr != _cdConstraintActive )
+        return *_cdConstraintActive;
+    else
+        return _constraintActive;
+}
+
+void ContextDependentPiecewiseLinearConstraint::registerBoundManager(
+    BoundManager *boundManager )
+{
+    ASSERT( nullptr == _boundManager );
```
guykatzz, line 59: we usually compare with the context on the right hand side

src/engine/ContextDependentPiecewiseLinearConstraint.cpp
```diff
@@ -0,0 +1,189 @@
...
+    ASSERT( nullptr != _context );
+    ASSERT( nullptr == _cdInfeasibleCases );
+    _cdInfeasibleCases = new ( true ) CVC4::context::CDList<PhaseStatus>( _context );
+}
+
+void ContextDependentPiecewiseLinearConstraint::initializeCDActiveStatus()
+{
+    ASSERT( nullptr != _context );
+    ASSERT( nullptr == _cdConstraintActive );
+    _cdConstraintActive = new ( true ) CVC4::context::CDO<bool>( _context, true );
```
guykatzz, line 85: shouldn't we initialize to _constraintActive?

src/engine/ContextDependentPiecewiseLinearConstraint.h
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /*
+       Mark that an exploredCase is infeasible, reducing the remaining search space.
+     */
+    void markInfeasible( PhaseStatus exploredCase );
+
+    /*
+      Returns phaseStatus representing next feasible case.
+      Returns PHASE_NOT_FIXED if no feasible case exists.
```
guykatzz, line 171: Can we perhaps add a dedicated sentinel? CONSTRAINT_INFEASIBLE or something?

src/engine/ContextDependentPiecewiseLinearConstraint.h
AleksandarZeljic: No, especially since the goal is to merge the two classes eventually. Fixed.

src/engine/ContextDependentPiecewiseLinearConstraint.h
AleksandarZeljic: Fixed.

src/engine/ContextDependentPiecewiseLinearConstraint.cpp
AleksandarZeljic: Fixed

src/engine/ContextDependentPiecewiseLinearConstraint.cpp
AleksandarZeljic: Fixed

src/engine/ContextDependentPiecewiseLinearConstraint.h
AleksandarZeljic: Fixed.

src/engine/ContextDependentPiecewiseLinearConstraint.h
AleksandarZeljic: So I would keep the name as is because there is a distinction in the return type - get*Case(s) all return PhaseStatus enumeration in some form, while this one actually returns an actual PiecewiseLinearCaseSplit. In the long run this should be the only method to return a PiecewiseLinearCaseSplit.

src/engine/ContextDependentPiecewiseLinearConstraint.h
AleksandarZeljic: Fixed.

src/engine/ContextDependentPiecewiseLinearConstraint.h
AleksandarZeljic: Fixed

src/engine/ContextDependentPiecewiseLinearConstraint.h
AleksandarZeljic: Fixed, changed style. The idea is to avoid accidental assignment, but I understand that with our style and reviews this should be unneccessary.

src/engine/ContextDependentPiecewiseLinearConstraint.h
AleksandarZeljic: I have not done that because it would make dealing with Max and Disjunction trickier, but I will see what I can do.

AleksandarZeljic: I was overthinking things, I added a high-valued entry to PhaseStatus enumeration, and updated this method accordingly.

AleksandarZeljic:
@guykatzz I think I've addressed all the concerns you raised, assuming the check passes, you can take another look. And if everything is fine I will rebase the branch onto master for a merge. Wasn't sure if rebase and force push would lose the review discussion or not.

src/engine/ContextDependentPiecewiseLinearConstraint.cpp
```diff
@@ -0,0 +1,189 @@
...
+    ASSERT( _cdConstraintActive == nullptr );
+    _cdConstraintActive = new ( true ) CVC4::context::CDO<bool>( _context, _constraintActive );
+}
+
+void ContextDependentPiecewiseLinearConstraint::initializeCDPhaseStatus()
+{
+    ASSERT( _context != nullptr );
+    ASSERT( _cdPhaseStatus == nullptr );
+    _cdPhaseStatus =
+        new ( true ) CVC4::context::CDO<PhaseStatus>( _context, PHASE_NOT_FIXED );
```
guykatzz, line 93: initialize to _phaseStatus?

guykatzz (approved):
LGTM (but see my one comment)

PR #429, opened 2021-01-30
Bound Manager Class

AleksandarZeljic:
BoundManager class is a context-dependent implementation of a centralized
variable registry and their bounds. The intent it so use a single BoundManager
object between multiple bound tightener classes, which enables
those classes to care only about bounds and forget about book-keeping. 
Furthermore, the context-dependent implementation eliminates the need for 
eager memory book-keeping.

BoundManager provides a method to obtain a new variable with:
 - registerNewVariable().

The bound values and tighten flags are stored using context-dependent objects,
which backtrack automatically with the central _context object.

There are two sets of methods to set bounds:
-  setLower/UpperBounds - local methods used to update bounds
-  tightenLower/Upper Bounds - shared method to update bounds, 
   propagates the new bounds to the _tableau (if registered) to keep the 
   assignment and basic/non-basic variables updated accordingly.
 
As soon as bounds become inconsistent, i.e. lowerBound > upperBound, an
InfeasableQueryException is thrown. In the long run, we want the exception
replaced by a flag, and switch to the conflict analysis mode instead.

It is assumed that variables are not introduced on the fly, and as such
interaction with context-dependent features is not implemented.

Another thing we may want to consider in the long run, is to make the 
BoundManager a VariableWatcher, and move those features from Tableau.

Integration into the master can be gradual, first allowing each class storing 
bounds a private copy of the object, before switching to a centralized object,
instant bound visibility and elimination of tightening propagation.

### DIFF
--- a/src/engine/BoundManager.cpp
+++ b/src/engine/BoundManager.cpp
@@ -43,20 +43,20 @@ BoundManager::~BoundManager()
 
 void BoundManager::initialize( unsigned numberOfVariables )
 {
-    ASSERT( 0 == _size );
+    ASSERT( _size == 0 );
 
     for ( unsigned i = 0; i < numberOfVariables; ++i )
         registerNewVariable();
 
-    ASSERT( numberOfVariables == _size );
+    ASSERT( _size == numberOfVariables );
 }
 
 unsigned BoundManager::registerNewVariable()
 {
-    ASSERT( _lowerBounds.size() == _size );
-    ASSERT( _upperBounds.size() == _size );
-    ASSERT( _tightenedLower.size() == _size );
-    ASSERT( _tightenedUpper.size() == _size );
+    ASSERT( _size == _lowerBounds.size() );
+    ASSERT( _size == _upperBounds.size() );
+    ASSERT( _size == _tightenedLower.size() );
+    ASSERT( _size == _tightenedUpper.size() );
 
     unsigned newVar = _size++;
 
@@ -81,7 +81,7 @@ unsigned BoundManager::getNumberOfVariables() const
 bool BoundManager::tightenLowerBound( unsigned variable, double value )
 {
     bool tightened = setLowerBound( variable, value );
-    if ( tightened && nullptr != _tableau )
+    if ( tightened && _tableau != nullptr )
         _tableau->ensureNonBasicVariableGTLB( variable, value );
     return tightened;
 }
@@ -89,7 +89,7 @@ bool BoundManager::tightenLowerBound( unsigned variable, double value )
 bool BoundManager::tightenUpperBound( unsigned variable, double value )
 {
     bool tightened = setUpperBound( variable, value );
-    if ( tightened && nullptr != _tableau )
+    if ( tightened && _tableau != nullptr )
         _tableau->ensureNonBasicVariableLTUB( variable, value );
     return tightened;
 }
@@ -160,6 +160,6 @@ bool BoundManager::consistentBounds( unsigned variable )
 
 void BoundManager::registerTableauReference( Tableau *ptrTableau )
 {
-    ASSERT( nullptr == _tableau );
+    ASSERT( _tableau == nullptr );
     _tableau = ptrTableau;
 }
--- a/src/engine/BoundManager.cpp
+++ b/src/engine/BoundManager.cpp
@@ -18,16 +18,30 @@
 
 using namespace CVC4::context;
 
-BoundManager::BoundManager( unsigned numberOfVariables, Context &context )
-    : _size( numberOfVariables )
+BoundManager::BoundManager( Context &context)
+    : _context(context)
+    , _size(0)
 {
-    for ( unsigned i = 0; i < numberOfVariables; ++i)
+};
+
+BoundManager::~BoundManager()
+{
+    for ( unsigned i = 0; i < _size; ++i)
     {
-      //CDList<double> lb ( &context, false );
-      //CDList<double> ub ( &context, false );
-      //new (true) CDList<double>( &context )
-      _lowerBounds.append( new (true) CDList<double>( &context ) );
-      _upperBounds.append( new (true) CDList<double>( &context ) );
+        _lowerBounds[i]->deleteSelf();
+        _upperBounds[i]->deleteSelf();
+    }
+};
+
+void BoundManager::initialize( unsigned numberOfVariables)
+{
+    ASSERT( 0 == _size);
+
+    _size = numberOfVariables;
+    for ( unsigned i = 0; i < _size; ++i)
+    {
+        _lowerBounds.append( new (true) CDList<double>( &_context ) );
+        _upperBounds.append( new (true) CDList<double>( &_context ) );
     }
 }
 
@@ -70,7 +84,6 @@ unsigned BoundManager::getSize( )
 {
   return _size;
 }
-
 
 //
 // Local Variables:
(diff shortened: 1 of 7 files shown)


src/engine/Tableau.cpp
```diff
@@ -1704,6 +1704,47 @@ bool Tableau::allBoundsValid() const
...
+    }
+}
+
+
+// Ensure that non-basic variables are within bounds
+void Tableau::ensureNonBasicVariableLTUB( unsigned variable, double value )
+{
+    unsigned index = _variableToIndex[variable];
+    if ( !_basicVariables.exists( variable ) )
+        {
```
guykatzz, line 1733: Indentation is off, { should be under the "if"

src/engine/Tableau.cpp
```diff
@@ -1704,6 +1704,47 @@ bool Tableau::allBoundsValid() const
...
+            if ( FloatUtils::lt( value, _nonBasicAssignment[index] ) )
+                setNonBasicAssignment( variable, value, true );
+        }
+    else
+        {
+            // Recompute the status of an affected basic variable
```
guykatzz, line 1739: Why does this guarantee that the basic variable moves within bounds?

src/engine/BoundManager.h
```diff
@@ -0,0 +1,128 @@
...
+    double getUpperBound( unsigned variable );
+
+    /*
+       Obtain a list of all the bound updates since the last call to
+       getTightenings.
+     */
+    void getTightenings( List<Tightening> &tightenings );
+
+    /*
+          Returns true if the bounds for the variable is valid, used to
```
guykatzz, line 96: indentation is off

src/engine/BoundManager.h
```diff
@@ -0,0 +1,128 @@
...
+          detect a conflict state.
+    */
+    bool consistentBounds( unsigned variable );
+
```
guykatzz, line 100: remove empty line

src/engine/BoundManager.h
```diff
@@ -0,0 +1,128 @@
...
+    /*
+       Silently sets bounds to the assigned value, checks bound consistency.
+     */
+    bool setLowerBound( unsigned variable, double value );
+    bool setUpperBound( unsigned variable, double value );
+
+    /*
+       Return current bound value.
+     */
+    double getLowerBound( unsigned variable );
```
guykatzz, line 86: all getters should be const

src/engine/BoundManager.h
```diff
@@ -0,0 +1,128 @@
...
+       Obtain a list of all the bound updates since the last call to
+       getTightenings.
+     */
+    void getTightenings( List<Tightening> &tightenings );
+
+    /*
+          Returns true if the bounds for the variable is valid, used to
+          detect a conflict state.
+    */
+    bool consistentBounds( unsigned variable );
```
guykatzz, line 99: make const

src/engine/BoundManager.h
```diff
@@ -0,0 +1,128 @@
...
+    unsigned _size;
+    Tableau *_tableau; // Used only by callbacks
+
+    Vector<CVC4::context::CDO<double> *> _lowerBounds;
+    Vector<CVC4::context::CDO<double> *> _upperBounds;
+
+    Vector<CVC4::context::CDO<bool> *> _tightenedLower;
+    Vector<CVC4::context::CDO<bool> *> _tightenedUpper;
+};
+
```
guykatzz, line 118: remove empty line

src/engine/BoundManager.h
```diff
@@ -0,0 +1,128 @@
...
+
+#endif // __BoundManager_h__
+
+//
```
guykatzz, line 122: this suffix can be removed, it became mostly irrelevant when we switched to cmake

src/engine/BoundManager.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+unsigned BoundManager::getNumberOfVariables()
+{
+    return _size;
+}
+
+bool BoundManager::tightenLowerBound( unsigned variable, double value )
+{
+    bool tightened = setLowerBound( variable, value );
+    if ( tightened &&  nullptr != _tableau )
```
guykatzz, line 83: delete space before nullptr, swap hands of comparison

src/engine/BoundManager.cpp
```diff
@@ -0,0 +1,172 @@
...
+        *_lowerBounds[variable] = value;
+        *_tightenedLower[variable] = true;
+        if ( !consistentBounds( variable ) )
+            throw InfeasibleQueryException();
+        return true;
+    }
+    return false;
+}
+
+bool BoundManager::setUpperBound( unsigned variable, double value )
```
guykatzz, line 110: This is not a problem, but an inconsistency with Tableau:
Traditionally, tighten() checks whether the new bound is tighter than the original, and only tightens in that case; whereas set just sets the bound, regardless of what was there before.

Set was typically used for initialization or for backtracking, when the bounds were actually made looser. 

I'm fine with keeping it as is, just wanted to point this out.

src/engine/Tableau.cpp
AleksandarZeljic: I don't think that it does - the method adjusts assignment of non-basic variables and recomputes the status of basic variables. I named it for the former part, but I can name it something more inclusive - updateVariableComplianceWithLower/UpperBound? For nonBasics that means updating assignment, for basics it means updating status. What are your thoughts?

Overall this is just an encapsulation of existing code to replicate current workflow when  bound updates come through the BoundManager.

src/engine/Tableau.cpp
AleksandarZeljic: Fixed

src/engine/BoundManager.h
AleksandarZeljic: I am not sure how to achieve this, since the bounds are stored within CDOs that are dynamically allocated, self-managed and need to be handled via non-constant  pointers. All straightforward attempts I made throw cascading errors inside ContextDependentObject class, which would be tricky (if at all possible) to resolve. I will think more about this and talk to the CVC4 people, but this seems likely to be a rabbit-hole.

src/engine/BoundManager.cpp
AleksandarZeljic: Thanks for pointing this out, it is good to keep in mind. 

The difference I was going for is that set local to BoundManager, whereas tighten is the polite method to be used by various sources of tighter bounds. 
 
It would be more efficient to check whether the bound is tighter before sending it to the bound manager anyway. However, I am skeptical of simply over-writing bounds though, because any bounds we set must progress the search. Backtracking is lazy and does not invoke any of these methods. Overall, It should be OK to move the check to tighten*, but I need to test it on the branch with all context-dependent stuff to ensure it does not break anything.

src/engine/BoundManager.h
AleksandarZeljic: Fixed, Vector::operator[] was the problem the entire time, switched it to get() and it complies.

src/engine/BoundManager.h
AleksandarZeljic: Fixed

src/engine/BoundManager.cpp
AleksandarZeljic: So upon reflecting, I think the implementation in the PR is more in line with how an SMT solver would handle a conflict. We arrive to a conflicted state from which conflict analysis repairs into a search state. Since we will be adding conflict analysis in the future, I will leave it like this.

src/engine/Tableau.cpp
guykatzz: Okay. Yes, that's fine, but please change the name to something more indicative.

guykatzz (approved):

src/engine/Tableau.cpp
AleksandarZeljic: Done.

PR #432, opened 2021-02-16
Update README.md

anwu1219:


AleksandarZeljic (commented):
This looks good. On the same topic,  the --help message still references DNC and DnC, those are quick search and replace and should be the last user facing occurrences of DNC.

anwu1219:
> This looks good. On the same topic, the --help message still references DNC and DnC, those are quick search and replace and should be the last user facing occurrences of DNC.

Updated the help message.

AleksandarZeljic (approved):
LGTM

PR #435, opened 2021-02-22
Fix deeppoly bug

anwu1219:
DeepPoly Analysis behaves incorrectly when a ReLU source node's neuron index does not match with ReLU output node's.
This PR fixes it, adds some new benchmarks and tests where the current master failed on.


guykatzz (approved):

guykatzz:
Does this happen for the other activation functions? Sign, Abs, Max, etc?

anwu1219:
> Does this happen for the other activation functions? Sign, Abs, Max, etc?

Max should be fine. But Sign and Abs have the same issue as ReLU. I've pushed the change and will merge once the tests pass.

PR #439, opened 2021-03-02
Add simulation for preprocess.

tagomaru:
This PR is going to add simulation function to enhance preprocess.


src/engine/Engine.cpp
```diff
@@ -1821,6 +1824,27 @@ List<unsigned> Engine::getInputVariables() const
     return _preprocessedQuery.getInputVariables();
 }
 
+void Engine::performSimulation()
+{
+    // outer vector is for neuron
+    // inner vector is for simulation value
+    std::vector<std::vector<double>> simulations;
```
anwu1219, line 1831: Could you try to use the Vector class instead of std::vector to be consistent with the rest of the project?
Hopefully a search and replace would do.

anwu1219: Like in https://github.com/NeuralNetworkVerification/Marabou/blob/8d777410f270af3bfdca747ed7f5acf7f3feed53/src/engine/BoundManager.h#L41

Vector is just a wrapper class around std::vector

src/engine/Engine.cpp
```diff
@@ -1821,6 +1824,27 @@ List<unsigned> Engine::getInputVariables() const
...
+
+    std::mt19937 mt( GlobalConfiguration::SIMULATION_RANDOM_SEED );
+
+    for ( unsigned i = 0; i < _networkLevelReasoner->getLayer( 0 )->getSize(); ++i )
+    {
+        std::uniform_real_distribution<double> distribution( _networkLevelReasoner->getLayer( 0 )->getLb( i ),
+                                                                _networkLevelReasoner->getLayer( 0 )->getUb( i ) ); 
+        std::vector<double> simulationInput( Options::get()->getInt( Options::NUMBER_OF_SIMULATIONS ) );
```
anwu1219, line 1839: Store NUMBER_OF_SOMULATIONS as a field variable in the constructor. Since parallel mode can call processInputQuery() in parallel, and the Options object is not thread safe.

src/engine/Engine.cpp
```diff
@@ -1821,6 +1824,27 @@ List<unsigned> Engine::getInputVariables() const
     return _preprocessedQuery.getInputVariables();
 }
 
+void Engine::performSimulation()
+{
+    // outer vector is for neuron
```
anwu1219, line 1829: Add a check to return directly if the number of simulation is 0

src/nlr/LPFormulator.cpp
```diff
@@ -279,10 +279,19 @@ void LPFormulator::optimizeBoundsWithLpRelaxation( const Map<unsigned, Layer *>
 
     gurobiStart = TimeUtils::sampleMicro();
 
+    bool skipTightenLb = false; // If true, skip lower bound tightening
+    bool skipTightenUb = false; // If true, skip upper bound tightening
+
+    unsigned skipTighterLowerBoundCounter = 0;
```
anwu1219, line 285: Since these two variables are used only for printing, maybe we should discard them.

anwu1219 (changes requested):

src/engine/Engine.cpp
tagomaru: done.

src/engine/Engine.cpp
tagomaru: done.

src/nlr/LPFormulator.cpp
tagomaru: I changed logging to print not counter but which neuron is skipped.

anwu1219 (approved):

src/nlr/LayerOwner.h
```diff
@@ -17,6 +17,9 @@
 #define __LayerOwner_h__
 
 #include "Tightening.h"
+#include "Vector.h"
+
+// #include <vector>
```
anwu1219, line 22: remove this line?

src/configuration/Options.cpp
```diff
@@ -58,6 +58,7 @@ void Options::initializeDefaultValues()
     _intOptions[VERBOSITY] = 2;
     _intOptions[TIMEOUT] = 0;
     _intOptions[CONSTRAINT_VIOLATION_THRESHOLD] = 20;
+    _intOptions[NUMBER_OF_SIMULATIONS] = 10;
```
anwu1219, line 61: Maybe set this to 1 for now? We can submit another PR after evaluating what is the best number of simulations.

src/nlr/LayerOwner.h
```diff
@@ -32,6 +35,7 @@ class LayerOwner
     virtual const ITableau *getTableau() const = 0;
     virtual unsigned getNumberOfLayers() const = 0;
     virtual void receiveTighterBound( Tightening tightening ) = 0;
+    virtual void simulate( Vector<Vector<double>> *input ) = 0;
```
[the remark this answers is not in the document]
tagomaru: done.

src/nlr/LayerOwner.h
tagomaru: done.

src/configuration/Options.cpp
tagomaru: done.

anwu1219 (approved):

PR #442, opened 2021-03-18
Inherit ContextDependentPiecewiseLinearConstraint in *Constraint classes

AleksandarZeljic:
This PR changes ReLU, AbsolutValue, Sign, Max and Disjunction Constraints 
to inherit ContextDependentPiecewiseLinearConstraint class, which prepares
them for transition to the new architecture.

- Implements virtual methods of ContextDependentPiecewiseLinearConstrain.
- Adds basic tests for context dependent behavior of all the constraint classes.
- Minor changes in Max and Disjunction constraint to use PhaseStatus enum.

Co-authored-by: ahmed-irfan <irfan@cs.stanford.edu>
Co-authored-by: anwu1219 <haozewu@stanford.edu>


src/engine/DisjunctionConstraint.cpp
```diff
@@ -19,9 +19,24 @@
 #include "Statistics.h"
 
 DisjunctionConstraint::DisjunctionConstraint( const List<PiecewiseLinearCaseSplit> &disjuncts )
-    : _disjuncts( disjuncts )
-    , _feasibleDisjuncts( disjuncts )
+    : ContextDependentPiecewiseLinearConstraint( disjuncts.size() )
+    , _disjuncts( disjuncts.begin(), disjuncts.end() )
+    , _feasibleDisjuncts( disjuncts.size(), 0 )
```
guykatzz, line 24: why not just _feasibleDisjuncts( disjuncts )? Instead of the loop in the body of the function

src/engine/DisjunctionConstraint.cpp
```diff
@@ -19,9 +19,24 @@
...
+    for ( unsigned ind = 0;  ind < disjuncts.size();  ++ind )
+        _feasibleDisjuncts.append( ind );
+
+    extractParticipatingVariables();
+}
+
+DisjunctionConstraint::DisjunctionConstraint( const Vector<PiecewiseLinearCaseSplit> &disjuncts )
+    : ContextDependentPiecewiseLinearConstraint( disjuncts.size() )
+    , _disjuncts( disjuncts )
+    , _feasibleDisjuncts( disjuncts.size(), 0 )
```
guykatzz, line 35: Same as above

src/engine/DisjunctionConstraint.cpp
```diff
@@ -127,17 +150,35 @@ List<PiecewiseLinearConstraint::Fix> DisjunctionConstraint::getSmartFixes( ITabl
 
 List<PiecewiseLinearCaseSplit> DisjunctionConstraint::getCaseSplits() const
 {
-    return _disjuncts;
+    return List<PiecewiseLinearCaseSplit>( _disjuncts.begin(), _disjuncts.end() );
```
guykatzz, line 153: What's the difference between the line erased and the new line?

src/engine/DisjunctionConstraint.h
```diff
@@ -174,15 +201,22 @@ class DisjunctionConstraint : public PiecewiseLinearConstraint
...
+        return static_cast<PhaseStatus>( ind + 1 );
+    }
+
+    inline unsigned phaseStatusToInd( PhaseStatus phase ) const
+    {
+        //ASSERT( phase != PHASE_NOT_FIXED );
+        return static_cast<unsigned>( phase ) - 1;
+    }
+
+
```
guykatzz, line 218: remove empty lines

src/engine/MaxConstraint.cpp
```diff
@@ -196,7 +221,16 @@ void MaxConstraint::notifyUpperBound( unsigned variable, double value )
...
-        _elements.erase( variable );
+    {
+        if ( _cdInfeasibleCases )
+            markInfeasible( variableToPhase( variable ) );
+        else
+            _elements.erase( variable );
+    }
+
+    // If all elements have been eliminated, set the phase of the constraint.
+    if ( _elements.size() == 0)
```
guykatzz, line 232: use _elements.empty()

src/engine/MaxConstraint.h
```diff
@@ -34,137 +46,197 @@ class MaxConstraint : public PiecewiseLinearConstraint
...
-    String serializeToString() const;
+    String serializeToString() const override;
 
     /*
      * Returns a boolean value indicating if at least one input variable was eliminated (True)
      * or not (False)
      */
     bool wereVariablesEliminated() const;
 
+
```
guykatzz, line 174: remove empty lines

src/engine/MaxConstraint.h
```diff
@@ -34,137 +46,197 @@ class MaxConstraint : public PiecewiseLinearConstraint
...
     bool _obsolete;
     bool _eliminatedVariables;
     double _maxValueOfEliminated;
 
+    /*
+       Functions that abstract away _maxIndex and _maxIndexSet, and use the
+       refactored PhaseStatus to store this information.
+
+       Makes two assumptions:
+        - 0 is not a valid variable value, since it's reserved for PHASE_NOT_FIXED
```
guykatzz, line 193: 0 is actually almost always a variable. can we get rid of this assumption?

src/engine/MaxConstraint.h
```diff
@@ -34,137 +46,197 @@ class MaxConstraint : public PiecewiseLinearConstraint
...
+        - the total number of variables does not reach the value of MAX_PHASE_ELIMINATED
+     */
+    inline bool maxIndexSet() const
+    {
+        return getPhaseStatus() != PHASE_NOT_FIXED;
```
guykatzz, line 198: I realized I don't fully understand the semantics of _phaseStatus and _cdPhaseStatus. Is this an indication of the phase that a non-fixed constraint is in? For MAX this makes sense (indicates the current element with maximal assignment), but what does it mean for Relu and Abs?

src/engine/MaxConstraint.h
```diff
@@ -34,137 +46,197 @@ class MaxConstraint : public PiecewiseLinearConstraint
...
+    inline PhaseStatus variableToPhase( unsigned variable ) const
+    {
+        return static_cast<PhaseStatus>( variable );
+    }
+
+    inline unsigned phaseToVariable( PhaseStatus phase ) const
+    {
+        return static_cast<unsigned>( phase );
+    }
+
```
guykatzz, line 237: Remove empty lines

src/engine/ReluConstraint.cpp
```diff
@@ -31,7 +33,8 @@
 #endif
 
 ReluConstraint::ReluConstraint( unsigned b, unsigned f )
-    : _b( b )
+    : ContextDependentPiecewiseLinearConstraint( 2u )
```
guykatzz, line 36: Use constant

src/engine/SignConstraint.cpp
```diff
@@ -27,7 +29,8 @@
 #endif
 
 SignConstraint::SignConstraint( unsigned b, unsigned f )
-    : _b( b )
+    : ContextDependentPiecewiseLinearConstraint( 2u )
```
guykatzz, line 32: Constant

guykatzz (commented):
Most of my comments are cosmetic. This is a pretty major change - have you run all available benchmarks to see that nothing breaks?

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -71,17 +73,25 @@ PiecewiseLinearFunctionType AbsoluteValueConstraint::getType() const
     return PiecewiseLinearFunctionType::ABSOLUTE_VALUE;
 }
 
-PiecewiseLinearConstraint *AbsoluteValueConstraint::duplicateConstraint() const
+ContextDependentPiecewiseLinearConstraint *AbsoluteValueConstraint::duplicateConstraint() const
 {
     AbsoluteValueConstraint *clone = new AbsoluteValueConstraint( _b, _f );
     *clone = *this;
+    this->initializeDuplicatesCDOs( clone );
```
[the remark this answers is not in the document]
AleksandarZeljic: Fixed

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -329,6 +339,21 @@ List<PiecewiseLinearCaseSplit> AbsoluteValueConstraint::getCaseSplits() const
     return splits;
 }
 
+List<PhaseStatus> AbsoluteValueConstraint::getAllCases() const
+{
+    return { ABS_PHASE_NEGATIVE, ABS_PHASE_POSITIVE};
```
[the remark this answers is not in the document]
AleksandarZeljic: PhaseStatus is a shared ENUM between all ConstraintClasses. I can add the constant with some generic name - BIPHASIC_PIECEWISE_LINEAR_CONSTRAINT. Although, it would make sense to have it as standalone constant to avoid mixing up status and number of cases. And it would be only constant of its kind, since all constraints so far are either fixed biphasic or have dynamic number of cases. What do you think?

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -378,7 +403,7 @@ bool AbsoluteValueConstraint::phaseFixed() const
     return _phaseStatus != PhaseStatus::PHASE_NOT_FIXED;
 }
 
-PiecewiseLinearCaseSplit AbsoluteValueConstraint::getValidCaseSplit() const
+PiecewiseLinearCaseSplit AbsoluteValueConstraint::getImpliedCaseSplit() const
```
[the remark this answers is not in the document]
AleksandarZeljic: As we move to a new SMT engine, we will switch to terms decision and impliedCase, which will return PhaseStatus cases and be used to retrieve the CaseSplit itself. So this method will go away soon anyway, I renamed it to indicate that they serve the same purpose.

src/engine/AbsoluteValueConstraint.h
```diff
@@ -2,23 +2,33 @@
...
- ** [[ Add lengthier description here ]]
-
+ ** AbsoluteValueConstraint implements the following constraint:
+ ** f = Abs( b )
+ **
+ ** It distinguishes two relevant phases for search:
+ ** ABS_PHASE_POSITIVE: b > 0
+ ** ABS_PHASE_NEGATIVE: b <=0
+ **
+ ** The constraint operates in two modes pre-processing mode, which is stores
```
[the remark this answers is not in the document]
AleksandarZeljic: Fixed.

src/engine/DisjunctionConstraint.cpp
AleksandarZeljic: The data structure differ now, it used to be a List and now its a Vector to utilize indexing using PhaseStatus. I didn't want to meddle with outer interface for now, but if you want I can go fix that where needed.

src/engine/DisjunctionConstraint.cpp
AleksandarZeljic: Return type _disjuncts is a Vector now. This method will become obsolete in the long run.

src/engine/MaxConstraint.h
AleksandarZeljic: Yeah, I introduced the conversion functions since, it is simple to just increment the stored value by one when reading/writing.

src/engine/MaxConstraint.h
AleksandarZeljic: In my mind the semantics is the same, the only difference is when we can act on the information. ReLU and other two-phase constraints do not depend on any additional variables, so we can treat them as implications immediately. Max is trickier because we need to convince ourselves that no other variable can exceed the value, before we can treat it as an implication.

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -71,17 +73,25 @@ PiecewiseLinearFunctionType AbsoluteValueConstraint::getType() const
...
     *clone = *this;
+    this->initializeDuplicatesCDOs( clone );
     return clone;
 }
 
 void AbsoluteValueConstraint::restoreState( const PiecewiseLinearConstraint *state )
 {
     const AbsoluteValueConstraint *abs = dynamic_cast<const AbsoluteValueConstraint *>( state );
+
+    CVC4::context::CDO<bool> *activeStatus = _cdConstraintActive;
```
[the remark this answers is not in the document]
AleksandarZeljic: So this is not fantastic logic, it is there simply to avoid losing context objects and causing memory leaks. Once we switch to context-dependent architecture the restoreState methods all over the place will be obsolete.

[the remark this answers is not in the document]
AleksandarZeljic: I can re-initialize their values to whatever we are restoring to, but it is really just paranoid code and not likely to be relevant in practice. Alternatively, I can replace current logic with assertions to indicate that the constraint is not in context-dependent mode.

src/engine/MaxConstraint.h
guykatzz: Sorry, I still don't follow. If max's phaseStatus() is x1, does this mean that currently x1 has the highest assignment?
Are we "overloading" this field to indicate this?

AleksandarZeljic: Yes, x1 would have the highest assignment, or higher lower bound than all the other upper bounfs. 

I don't think the field is overloaded. 
Relu = max ( x, 0) vs Max = max( x, y). 

Relu can propagate as soon as we know something about x since zero is a constant. But max(x,y) can propagate only once it also knows that the upper bound of y is less than the value of x, otherwise it is not sufficient information. 

Basically, _phaseStatus can be thought of as the phase under the current bounds (and implicitly assignment). Relu needs information over one variable for propagation, but max needs also to know bounds on all the other variables involved to propagate. _phaseStatus denotes the same, the difference is that ReLU never has to wait on other variables so there is no period of uncertainty. Does this make sense?

AleksandarZeljic:
@guykatzz I had to re-base the branch, can you approve again? There was one bug fix since your last review.

PR #446, opened 2021-04-02
Split & Conquer guide

AleksandarZeljic:
This PR adds a Jupyter notebook that illustrates the use of Split & Conquer mode.

The PR also refactors some outdated names for parameters of maraboupy.


anwu1219 (approved):
LGTM! Thanks for doing this.

PR #452, opened 2021-05-04
LP Tightening after splits

tagomaru:
This PR is going to add a feature to call LP tightening after case splits.
The new LP tightening is done for only a first hidden layer and an output layer and accounts for every constraint of a whole network and updated bounds of split nodes.


src/configuration/GlobalConfiguration.cpp
```diff
@@ -92,6 +92,8 @@ const unsigned GlobalConfiguration::GUROBI_NUMBER_OF_THREADS = 1;
 const bool GlobalConfiguration::GUROBI_LOGGING = false;
 #endif // ENABLE_GUROBI
 
+const bool GlobalConfiguration::LP_TIGHTENING_AFTER_SPLIT = true;
```
anwu1219, line 95: make this a command-line argument and (also add it to the python API)

src/engine/Engine.cpp
```diff
@@ -1192,6 +1200,42 @@ void Engine::performMILPSolverBoundedTightening()
...
+
+        switch ( Options::get()->getMILPSolverBoundTighteningType() )
+        {
+        case MILPSolverBoundTighteningType::LP_RELAXATION:
+        case MILPSolverBoundTighteningType::LP_RELAXATION_INCREMENTAL:
+            _networkLevelReasoner->lpRelaxation( targetIndex );
+            break;
+
+        case MILPSolverBoundTighteningType::MILP_ENCODING:
+        case MILPSolverBoundTighteningType::MILP_ENCODING_INCREMENTAL:
```
anwu1219, line 1217: It looks like MILP_ENCODING_INCREMENTAL and LP_RELAXATION_INCREMENTAL are not implemented? Maybe also return in those cases?

tagomaru: I changed so that it just returns.

src/nlr/LPFormulator.h
```diff
@@ -81,6 +84,49 @@ class LPFormulator : public ParallelSolver
     bool _cutoffInUse;
     double _cutoffValue;
 
+    /*
```
anwu1219, line 87: Why can't the same thread argument be reused?

src/nlr/NetworkLevelReasoner.h
```diff
@@ -117,7 +117,9 @@ class NetworkLevelReasoner : public LayerOwner
     void symbolicBoundPropagation();
     void deepPolyPropagation();
     void lpRelaxationPropagation();
+    void LPTigheningForOneLayer( unsigned targetIndex );
```
anwu1219, line 120: Tighening -> Tightening

src/nlr/NetworkLevelReasoner.h
```diff
@@ -117,7 +117,9 @@ class NetworkLevelReasoner : public LayerOwner
...
     void MILPPropagation();
+    void MILPTigheningForOneLayer( unsigned targetIndex );
```
anwu1219, line 122: Tighening -> Tightening

src/engine/Engine.cpp
```diff
@@ -1192,6 +1200,44 @@ void Engine::performMILPSolverBoundedTightening()
     }
 }
 
+void Engine::performMILPSolverBoundedTighteningForSingleLayer( unsigned targetIndex )
+{
+    if ( _networkLevelReasoner && Options::get()->gurobiEnabled() && !Options::get()->getBool( Options::SKIP_LP_TIGHTENING_AFTER_SPLIT ) )
```
anwu1219, line 1205: Add another check condition
Options::get()->getMILPSolverBoundTighteningType() != MILPSolverBoundTighteningType::NONE

anwu1219: Since the Options object is not thread safe and there is a chance that multiple Engine object be accessing the Options object.  Maybe store  `Options::get()->gurobiEnabled(),  Options::get()->getBool( Options::SKIP_LP_TIGHTENING_AFTER_SPLIT ) and Options::get()->getMILPSolverBoundTighteningType()` as local variables in the constructor of Engine()

tagomaru: Done.

src/nlr/NetworkLevelReasoner.h
tagomaru: Done.

src/nlr/LPFormulator.h
tagomaru: Reused now.

src/configuration/GlobalConfiguration.cpp
tagomaru: Done.

src/engine/Engine.cpp
```diff
@@ -1192,6 +1202,45 @@ void Engine::performMILPSolverBoundedTightening()
...
 }
 
+void Engine::performMILPSolverBoundedTighteningForSingleLayer( unsigned targetIndex )
+{
+    if ( _networkLevelReasoner && _isGurobyEnabled && !_isSkipLpTighteningAfterSplit
+            && _milpSolverBoundTighteningType != MILPSolverBoundTighteningType::NONE )
+    {
+        _networkLevelReasoner->obtainCurrentBounds();
+
+        switch ( Options::get()->getMILPSolverBoundTighteningType() )
```
anwu1219, line 1212: use _milpSolverBoundTighteningType

anwu1219 (approved):

src/engine/Engine.cpp
tagomaru: Done.

PR #487, opened 2021-11-06
Context-dependent SmtCore

AleksandarZeljic:
This PR introduces a class with context-dependent implementation of the SmtCore. 

The CDSmtCore class implementation relies on:
- Access to the unique context object from which all the context-dependent structures are obtained.
- PiecewiseLinearConstraint class stores its search state in a context-dependent manner (already implemented) and 
modifying it using nextFeasibleCase()/markInfeasible() methods.
- Using BoundManager class to store bounds in a context-dependent manner (to be integrated in a future patch).

The CDSmtCore distinguishes between: **decisions** and **implications**. 
Decision is a case of PiecewiseLinearConstraint (represented using PhaseStatus enum), which is being asserted on the trail. The fact that it is a decision implies that the PiecewiseLinearConstraint has at least one other feasible case.
Decisions represent nodes/search-states in the search-space tree.

Implication is a case of PiecewiseLinearConstraint (represented using PhaseStatus enum), which is being asserted on the trail. Implication represents the last feasible case of the PiecewiseLinearConstraint.

_trail is a chronological list of cases of PiecewiseLinearConstraints asserted to hold (represented using TrailEntry) - both decisions and implications.
_decisions is a chronological list of decisions stored on the trail. 
_trail and _decisions are both context dependent and will synchronize in unison with the context object.

When we split on cases of a PiecewiseLinearConstraint a decision is performed.
When a case of PiecewiseLinearConstraint is fixed via bound propagation an implication is performed.
When a search state is found to be infeasible, CDSmtCore backtracks lazily to the last decision and continues the search.

Context management is done automatically when a new decision is pushed and a decision level is popped.
Popping a decision will cause all related context dependent data structures to backtrack in sync with the _trail.
This will effectively backtrack the entire search state via PiecewiseLinearConstraints and boundManager. 
The only exception is the state of basic/non-basic variables in the tableau, which may need to be recalculated. 
However, no additional memory overhead is incurred in this process.

Next steps: Integration of the centralized context object, proper initialization of all context dependent structures and integrating them all through via the Engine and CDSmtCore class.

### DIFF
--- a/src/engine/CDSmtCore.cpp
+++ b/src/engine/CDSmtCore.cpp
@@ -302,7 +302,7 @@ void CDSmtCore::interruptIfCompliantWithDebugSolution()
 }
 
 
-PiecewiseLinearCaseSplit CDSmtCore::getDecision( unsigned decisionLevel )
+PiecewiseLinearCaseSplit CDSmtCore::getDecision( unsigned decisionLevel ) const
 {
     ASSERT( decisionLevel <= getDecisionLevel() );
     ASSERT( decisionLevel > 0 );
--- a/src/engine/CDSmtCore.cpp
+++ b/src/engine/CDSmtCore.cpp
@@ -2,7 +2,7 @@
 /*! \file CDSmtCore.cpp
  ** \verbatim
  ** Top contributors (to current version):
- **   Guy Katz, Aleksandar Zeljic,  Parth Shah
+ **   Guy Katz, Aleksandar Zeljic, Parth Shah
  ** This file is part of the Marabou project.
  ** Copyright (c) 2017-2019 by the authors listed in the file AUTHORS
  ** in the top-level source directory) and their institutional affiliations.
@@ -131,7 +131,7 @@ void CDSmtCore::decide()
     decideSplit( _constraintForSplitting );
 }
 
-void CDSmtCore::decideSplit( PiecewiseLinearConstraint * constraint )
+void CDSmtCore::decideSplit( PiecewiseLinearConstraint *constraint )
 {
     struct timespec start = TimeUtils::sampleMicro();
 
--- a/src/engine/CDSmtCore.cpp
+++ b/src/engine/CDSmtCore.cpp
@@ -306,16 +306,9 @@ bool CDSmtCore::checkSkewFromDebuggingSolution()
         }
         else
         {
-            // If the active split is non-compliant but there are alternatives, that's fine
+            // If the active split is non-compliant but there are alternatives, i.e. it was a decision, that's fine
             if ( isDecision && !splitAllowsStoredSolution( caseSplit, error ) )
             {
-                if ( !trailEntry.isFeasible() )
-                {
-                    printf( "Error! Have a split that is non-compliant with the stored solution, "
-                            "without alternatives:\n\t%s\n", error.ascii() );
-                    throw MarabouError( MarabouError::DEBUGGING_ERROR );
-                }
-
                 // Active split is non-compliant but this is fine, because there are alternatives. We're done.
                 return false;
             }
--- a/src/engine/CDSmtCore.cpp
+++ b/src/engine/CDSmtCore.cpp
@@ -1,0 +1,541 @@
+/*********************                                                        */
+/*! \file CDSmtCore.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Aleksandar Zeljic, Guy Katz, Parth Shah, Duligur Ibeling
+ ** This file is part of the Marabou project.
+ ** Copyright (c) 2017-2019 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved. See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** The CDSmtCore implements a CDCL-style loop, using context-dependent data types
+ ** and state-based PiecewiseLinearConstraint to store the search state in an
+ ** incremental and easy-to-backtrack way.
+
+ **/
+
+#include "Debug.h"
+#include "DivideStrategy.h"
+#include "EngineState.h"
+#include "FloatUtils.h"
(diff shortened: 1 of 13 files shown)


src/engine/Engine.h
```diff
@@ -126,6 +126,11 @@ class Engine : public IEngine, public SignalHandler::Signalable
     */
     void applySplit( const PiecewiseLinearCaseSplit &split );
 
+    /*
+      Recompute status of basic variables
+    */
+    void recomputeBasicStatus() { _tableau->computeBasicStatus(); };
```
guykatzz, line 132: This doesn't seem like a method that should appear in the Engine.
Perhaps we should only expose a Tableau method called "contextPopHook", or something like that, that will get invoked after a pop?

src/engine/TrailEntry.h
```diff
@@ -0,0 +1,61 @@
...
+  {
+      _pwlConstraint->markInfeasible( _phase );
+  }
+
+  inline bool isFeasible() const
+  {
+      return _pwlConstraint->isFeasible();
+  }
+
+ TrailEntry(PiecewiseLinearConstraint * pwlc, PhaseStatus phase )
```
guykatzz, line 45: space after (

src/engine/CDSmtCore.h
```diff
@@ -0,0 +1,270 @@
+/*********************                                                        */
+/*! \file CDSmtCore.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Guy Katz, AleksandarZeljic,  Parth Shah
```
guykatzz, line 5: remove space before Parth

src/engine/CDSmtCore.h
```diff
@@ -0,0 +1,270 @@
...
+ ** to the last decision and continues the search.
+ **
+ ** Context management is done automatically when a new decision is pushed and a
+ ** decision level is popped. Popping a decision will cause all related context
+ ** dependent data structures to backtrack in sync with the _trail. This will
+ ** effectively backtrack the entire search state via PiecewiseLinearConstraints
+ ** and _boundManager. The only exception is the state of basic/non-basic
+ ** variables in the tableau, which may need to be recalculated. However, no
+ ** additional memory overhead is incurred in this process.
+ **
```
guykatzz, line 47: These comments are great, well done!

src/engine/CDSmtCore.h
```diff
@@ -0,0 +1,270 @@
...
+    void pushDecision( PiecewiseLinearConstraint *constraint,  PhaseStatus decision );
+
+    /*
+      Inform SmtCore of an implied (formerly valid) case split that was discovered.
+    */
+    void pushImplication( PiecewiseLinearConstraint *constraint );
+
+    /*
+        Pushes trail entry onto trail, handles decision book-keeping and
+        Update bounds and add equations to the engine.
```
guykatzz, line 122: Update --> update

src/engine/CDSmtCore.h
```diff
@@ -0,0 +1,270 @@
...
+    void decide();
+
+    /*
+      Decide a constraint's feasible case. Update bounds, add equations
+      and trail.
+     */
+    void decideSplit( PiecewiseLinearConstraint *constraint );
+
+    /*
+       Backtracks fully explored decisions and stores the last (feasible decision
```
guykatzz, line 138: remove (

src/engine/CDSmtCore.h
```diff
@@ -0,0 +1,270 @@
...
+    /*
+      For debugging purposes only
+    */
+    Map<unsigned, double> _debuggingSolution;
+
+   /*
+      Split when some relu has been violated for this many times
+    */
+    unsigned _constraintViolationThreshold;
+
```
guykatzz, line 266: remove empty line

src/engine/CDSmtCore.cpp
```diff
@@ -0,0 +1,432 @@
+/*********************                                                        */
+/*! \file CDSmtCore.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Guy Katz, Aleksandar Zeljic,  Parth Shah
```
guykatzz, line 5: Remove space before Parth

src/engine/CDSmtCore.cpp
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    ASSERT( _constraintForSplitting->isActive() );
+    _needToSplit = false;
+    _constraintForSplitting->setActiveConstraint( false );
+
+    decideSplit( _constraintForSplitting );
+}
+
+void CDSmtCore::decideSplit( PiecewiseLinearConstraint * constraint )
```
guykatzz, line 134: remove space after *

guykatzz (commented):
Hi Alex,
Looks very good. Most of my comments are cosmetic.

At various points you mention "lazily backtrack...". Where is this laziness expressed?

AleksandarZeljic:
Hi Guy, thanks for the comments, I'll fix them! 

All the context-dependent structures are inherently lazy. Backtracking updates the end pointer of the data structure to an earlier position - which effectively forgets everything beyond that point in 0(1) time. I can add a clarification to the comments.

guykatzz:
Okay. So laziness here just means that the actual freeing of the memory
happens at some later time?

AleksandarZeljic:
Yes, the memory is acquired by Context and it is only freed once at the very end. All the context-dependent data structures live in the Context's memory and their synchronized management is achieved via pointer updates.

AleksandarZeljic:
The key point of laziness here is that pointer update does not care about the size of memory being backtracked - whether it is one list element or a thousand the effort is the same.

src/engine/Engine.h
AleksandarZeljic: Great point, done!

AleksandarZeljic:
I've addressed all comments, @guykatzz, you can take another look. 

I've clarified the comments further to address the laziness, let me know if there is anything else that stands out.

guykatzz (approved):

Issue #481, opened 2021-09-29
Cmake .. returns errors


asidddique:
Hi, 

I am quite interested in using this tool. I tried the following commands after cloning the repository:

> cd path/to/marabou/repo/folder
> mkdir build 
> cd build
> cmake ..

However, the last command returns some error. It would be great if you could help in fixing this issue. The error list shown in the command window is given below:

> getarch_2nd.c:70:50: error: ‘DGEMM_DEFAULT_Q’ undeclared (first use in this function); did you mean ‘QGEMM_DEFAULT_UNROLL_N’?
>      printf("#define DLOCAL_BUFFER_SIZE\t%ld\n", (DGEMM_DEFAULT_Q * DGEMM_DEFAULT_UNROLL_N * 2 * 1 *  sizeof(double)));
>                                                   ^~~~~~~~~~~~~~~
>                                                   QGEMM_DEFAULT_UNROLL_N
> getarch_2nd.c:71:50: error: ‘CGEMM_DEFAULT_Q’ undeclared (first use in this function); did you mean ‘QGEMM_DEFAULT_UNROLL_N’?
>      printf("#define CLOCAL_BUFFER_SIZE\t%ld\n", (CGEMM_DEFAULT_Q * CGEMM_DEFAULT_UNROLL_N * 4 * 2 *  sizeof(float)));
>                                                   ^~~~~~~~~~~~~~~
>                                                   QGEMM_DEFAULT_UNROLL_N
> getarch_2nd.c:72:50: error: ‘ZGEMM_DEFAULT_Q’ undeclared (first use in this function); did you mean ‘QGEMM_DEFAULT_UNROLL_N’?
>      printf("#define ZLOCAL_BUFFER_SIZE\t%ld\n", (ZGEMM_DEFAULT_Q * ZGEMM_DEFAULT_UNROLL_N * 2 * 2 *  sizeof(double)));
>                                                   ^~~~~~~~~~~~~~~
>                                                   QGEMM_DEFAULT_UNROLL_N
> make: *** [Makefile.prebuild:66: getarch_2nd] Error 1
> Makefile.install:21: *** OpenBLAS: Please run "make" firstly.  Stop.
> make: *** [Makefile:340: install] Error 2
> -- Check if compiler accepts -pthread
> -- Check if compiler accepts -pthread - yes
> -- Found Threads: TRUE
[...]

PR #488, opened 2021-11-10
[configuration] Make cmake configuration fail early upon boost or openBLAS installation failure

lambdaxdotx:
**TLDR:** This PR introduces a couple of checks for the `cmake` configuration process to make it fail upon the installation failure of either `boost` or `openblas`.

## Problem

At the moment, whenever any `tools/download_*.sh` script fails, the configuration process continues w/o any explicit message, for failing only at the end. This may surprise and mislead the (non-dev) user: issue https://github.com/NeuralNetworkVerification/Marabou/issues/481 seems to be an example of that.

Catching the erratic state of the configuration and fail early seems a better strategy to me.

## Proposed solution

- Make `tools/download_*.sh` exit on any failure by passing the `-e` to `/bin/bash`
- Check the exit code of those scripts in the `CMakeLists.txt` and fail accordingly.

Let me know what you think.


lambdaxdotx:
The CI fails because the `download_boost.sh` fails at some point. In particular, this means that the script was _already_ failing.

The next step, I guess, is to look at the precise command that fails in the script and let maintainers decide how to proceed.

AleksandarZeljic:
Thanks for submitting the PR. I am pretty sure that the failure has nothing to do with the PR. A recent PR was pushed into master after passing CI checks, but once integrated failed because boost wasn't downloaded correctly only to suddenly work again without intervention. I will look into this.

lambdaxdotx:
Hi @AleksandarZeljic

Thanks for giving this PR some attention. Let me know if you need help.

PR #497, opened 2021-12-31
Add comments to property files

MatthewDaggitt:
I'm automatically generating Marabou query files, and it's useful to be able to put some debug information into them, so I've added the ability to have comments in property files. Any line which begins with `--` is ignored as a comment.

If the maintainers are happy with this addition, I would be grateful if someone could point me at suitable test files to insert some comments in. Unless I'm mistaken all the tests that run with `ctest -L unit` seem to be engine tests and don't involve the parser...


MatthewDaggitt:
Any chance at getting a review for this PR?

AleksandarZeljic (changes requested):
Hi Matthew,

This looks good to me. 

Can you make the comment characters into a pair of slashes "//" instead?  
And can you add comments to one or more of the property files used in regression tests? That would be an easy way to ensure this works.

Thanks in advance!

MatthewDaggitt:
Thanks! I've changed the comment token as requested and added comments to one of the mnist regression tests:
```
marabou_add_mnist_test(0 "mnist10x20.nnet" "image1_target1_epsilon0.005.txt" unsat)
```
Let me know if anything else needs doing :+1:

AleksandarZeljic (approved):
LGTM

MatthewDaggitt:
Thanks! 

I couldn't add it as part of a PR, but it might be worth adding a note about the comment token to the [wiki](https://github.com/NeuralNetworkVerification/Marabou/wiki/Marabou-Input-Formats)?

AleksandarZeljic:
Good point! Added comments to the grammar in the wiki.
________________________________

PR #500, opened 2022-01-15
Add optimization method in the Engine class

anwu1219:
1. Extend the Tableau and CostFunctionManager to handle optimization.
2. Add a minimizeHeuristicCost() method to the Engine class.

### DIFF
--- a/src/engine/CostFunctionManager.cpp
+++ b/src/engine/CostFunctionManager.cpp
@@ -152,8 +152,19 @@ void CostFunctionManager::computeGivenCostFunction( const Map<unsigned, double>
     computeMultipliers();
     computeReducedCosts();
 
-    // So that it doesn't count as "fresh" for declaring UNSAT
     _costFunctionStatus = ICostFunctionManager::COST_FUNCTION_JUST_COMPUTED;
+}
+
+double CostFunctionManager::computeGivenCostFunctionDirectly( const Map<unsigned, double>
+                                                              &heuristicCost )
+{
+    double cost = 0;
+    for ( const auto &pair : heuristicCost )
+        {
+            double value = _tableau->getValue( pair.first );
+            cost += pair.second * value;
+        }
+    return cost;
 }
 
 void CostFunctionManager::computeCoreCostFunction()
--- a/src/engine/CostFunctionManager.cpp
+++ b/src/engine/CostFunctionManager.cpp
@@ -117,6 +117,44 @@ void CostFunctionManager::computeCostFunction( const Map<unsigned, double> &heur
     computeReducedCosts();
 }
 
+void CostFunctionManager::computeGivenCostFunction( const Map<unsigned, double>
+                                                    &heuristicCost )
+{
+    /*
+      A heuristic-based cost function is computed by computing the core
+      cost function and adding to it the provided heuristic cost.
+
+      The heuristic cost may include variables that are basic and variables
+      that are non-basic. The basic variables are added to the vector of basic
+      costs, which is normally used in computing the core cost fuction.
+      Afterwards, once the modified core cost function has been computed,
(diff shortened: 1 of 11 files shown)


src/engine/CostFunctionManager.cpp
```diff
@@ -117,6 +117,45 @@ void CostFunctionManager::computeCostFunction( const Map<unsigned, double> &heur
...
+            _basicCosts[variableIndex] += cost;
+        else
+            _costFunction[variableIndex] += cost;
+    }
+
+    // Complete the calculation of the modified core cost function
+    computeMultipliers();
+    computeReducedCosts();
+
+    // So that it doesn't count as "fresh" for declaring UNSAT
```
guykatzz, line 155: Didnt understand this comment. If you set it to "Just computed", how doesn't it count as fresh?

src/engine/ITableau.h
```diff
@@ -180,6 +180,16 @@ class ITableau
     virtual bool areLinearlyDependent( unsigned x1, unsigned x2, double &coefficient, double &inverseCoefficient ) = 0;
     virtual unsigned getVariableAfterMerging( unsigned variable ) const = 0;
     virtual void postContextPopHook() = 0;
+
+    bool isOptimizing() const
+    {
+        return _optimizing;
+    }
+    void optimizing() { _optimizing = true; }
```
guykatzz, line 188: This is a bit inconsistent with the rest of the code. Can we switch to a single method, e.g., toggleOptimization( bool ), instead of having two setters?

src/engine/Engine.h
```diff
@@ -405,12 +416,14 @@ class Engine : public IEngine, public SignalHandler::Signalable
      */
     String _queryId;
 
+    Map<unsigned, double> _heuristicCost;
 
     /*
       Perform a simplex step: compute the cost function, pick the
       entering and leaving variables and perform a pivot.
+      Return true only if the current assignment is optimal.
```
guykatzz, line 424: optimal with respect to what?

src/engine/Engine.cpp
```diff
@@ -2455,3 +2464,74 @@ const Preprocessor *Engine::getPreprocessor()
...
+            continue;
+        }
+
+        localOptimaReached = performSimplexStep();
+    }
+    _tableau->notOptimizing();
+    ENGINE_LOG( "Optimizing w.r.t. the current heuristic cost - done\n" );
+}
+
+double Engine::computeHeuristicCost( const Map<unsigned, double> &heuristicCost )
```
guykatzz, line 2528: Should we perhaps move this functionality to the cost function manager?

guykatzz (commented):
Overall, looks good, except for minor comments.
This new functionality is not standard, text-book Simplex, so please be generous with comments describing the various functions :).

anwu1219:
> Overall, looks good, except for minor comments. This new functionality is not standard, text-book Simplex, so please be generous with comments describing the various functions :).

Sounds good! I'll add more comments in future PRs that involve SoI and MCMC-sampling.

src/engine/CostFunctionManager.cpp
anwu1219: Whoops sorry. I meant to say "so that the cost function is deemed "fresh" when no pivoting candidates are found. I removed this comment.

src/engine/ITableau.h
anwu1219: Done

src/engine/Engine.cpp
anwu1219: Done.

src/engine/Engine.h
anwu1219: With respect to _heuristicCost. I updated the comment.

guykatzz (approved):

PR #502, opened 2022-01-17
Extend ReLU/Max constraint with SoI cost term.

anwu1219:
Add methods in the ReLU constraints to contribute a cost term in the SoI function. This is a small PR. 
The next one is more involved with a SumOfInfeasibilitiesManager class which maintains/update the current heuristic cost (sum of the cost terms from the piecewise linear constraints).


src/engine/PiecewiseLinearConstraint.h
```diff
@@ -254,6 +255,37 @@ class PiecewiseLinearConstraint : public ITableau::VariableWatcher
...
+      "closer to satisfied" in the given phase status.
+    */
+    virtual void addCostFunctionComponent( Map<unsigned, double> &/* cost */,
+                                           PhaseStatus /* phase */ ) const {}
+
+    /*
+      Compute the error of the current assignment with respect to the cost compoenent
+      corresponding to the given phase.
+    */
+    virtual double computeCostFunctionComponent( PhaseStatus /* phase */ ) const
```
guykatzz, line 272: Due to their names, it sounds like {add/compute}CostFunctionComponent are doing similar things. But I guess compute is computing not the component, but its value?

src/engine/PiecewiseLinearConstraint.h
```diff
@@ -254,6 +255,37 @@ class PiecewiseLinearConstraint : public ITableau::VariableWatcher
...
+    }
+
+    /*
+      Ask the piecewise linear constraint to remove the cost term corresponding
+      to the given phase from the cost function.
+      The method does not check whether the cost term has been previously added.
+      This is done in the SumOfInfeasibilitiesManager class, which maintains/updates the
+      SoI function.
+    */
+    virtual void removeCostFunctionComponent( Map<unsigned, double> &/* cost */,
```
guykatzz, line 286: Instead of add/remove, why not have a get(), and let the manager do with it as it will?

guykatzz (commented):
I think it would be cleaner if the PLCs didn't manipulate the cost directly. Let them return their cost, and let someone else manipulate the actual array that stores it.

src/engine/PiecewiseLinearConstraint.h
anwu1219: Yeah exactly. add is adding a "symbolic term" to the cost function, compute returns the values of the term based on the current assignment.

src/engine/PiecewiseLinearConstraint.h
anwu1219: Yeah that makes sense. I'll defer this method to the SoIManager class.

anwu1219:
Updates:
I also added support of adding SoI terms from Max constraint.

src/engine/MaxConstraint.cpp
```diff
@@ -512,11 +512,10 @@ List<PiecewiseLinearConstraint::Fix> MaxConstraint::getSmartFixes( ITableau * )
 
 List<PhaseStatus> MaxConstraint::getAllCases() const
 {
-    if ( _phaseStatus != PHASE_NOT_FIXED )
+    if ( phaseFixed() )
```
anwu1219, line 515: This is a quick fix to the issue (https://github.com/NeuralNetworkVerification/Marabou/issues/504) resulting from the refactored MaxConstraint. For now this is probably enough for the current master to be sound, because most of the refactored functionality is not in use. A more thorough change in the MaxConstraint is needed for the context dependent stuff to be actually usable.

src/engine/LinearExpression.h
```diff
@@ -0,0 +1,47 @@
...
+ **
+ ** [[ Add lengthier description here ]]
+
+ **/
+
+#ifndef __LinearExpression_h__
+#define __LinearExpression_h__
+
+#include "Map.h"
+
```
guykatzz, line 20: Is this class different from the Equation class? Can we use the existing one?

src/engine/MaxConstraint.cpp
```diff
@@ -785,3 +832,21 @@ String MaxConstraint::serializeToString() const
 
     return output;
 }
+
```
guykatzz, line 835: remove empty line

src/engine/LinearExpression.h
anwu1219: The difference is that linear expression is the lhs of an equation plus a constant: A**x** +b

src/engine/MaxConstraint.cpp
anwu1219: Whoops will do

src/engine/LinearExpression.h
guykatzz: Okay. I believe an equation is Ax = b, so it could also be used, right?
Just trying to reduce code duplication

guykatzz: Never mind, I take my comment back. I think the linear expression has a sufficiently different meaning

src/engine/MaxConstraint.cpp
anwu1219: Done

guykatzz (approved):

PR #506, opened 2022-01-27
More regression tests

anwu1219:
1. Add add_test_input_query primitive, which allows regression testing beyond .nnet networks and ReLU
2. Add easy regression test on max pooling networks to regress1. 10 sat, 10 unsat. The unsat ones are quick and the sat ones take at most ~10 seconds.


AleksandarZeljic (approved):
LGTM.

PR #508, opened 2022-02-02
Update CMakeLists.txt to fix compilation error on recent MacOS versions.

anwu1219:
https://stackoverflow.com/questions/10046114/in-cmake-how-can-i-test-if-the-compiler-is-clang


AleksandarZeljic (approved):
LGTM

PR #509, opened 2022-02-02
export sdk path when installing OpenBLAS on Mac.

anwu1219:
Fixing the OpenBlas installation failure on Mac following this post: https://stackoverflow.com/questions/51761599/cannot-find-stdio-h


AleksandarZeljic (approved):
LGTM

PR #511, opened 2022-02-02
Access ExitCode from maraboupy 

anwu1219:
1. Now the solve method of Maraboupy returns the exitcode of the Engine. Before there was no way to differ "error" and "unsat".
2. Extended the tests/examples accordingly.
3. decrease the default value of the numerical tolerance.

### DIFF
--- a/src/configuration/GlobalConfiguration.cpp
+++ b/src/configuration/GlobalConfiguration.cpp
@@ -59,7 +59,7 @@ const double GlobalConfiguration::SYMBOLIC_TIGHTENING_ROUNDING_CONSTANT = 0.0000
 const bool GlobalConfiguration::PREPROCESS_INPUT_QUERY = true;
 const bool GlobalConfiguration::PREPROCESSOR_ELIMINATE_VARIABLES = true;
 const bool GlobalConfiguration::PREPROCESSOR_PL_CONSTRAINTS_ADD_AUX_EQUATIONS = true;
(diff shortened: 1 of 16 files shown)


src/configuration/GlobalConfiguration.cpp
```diff
@@ -70,8 +70,8 @@ const unsigned GlobalConfiguration::PSE_ITERATIONS_BEFORE_RESET = 1000;
 const double GlobalConfiguration::PSE_GAMMA_ERROR_THRESHOLD = 0.001;
 const double GlobalConfiguration::PSE_GAMMA_UPDATE_TOLERANCE = 0.000000001;
 
-const double GlobalConfiguration::RELU_CONSTRAINT_COMPARISON_TOLERANCE = 0.00001;
-const double GlobalConfiguration::ABS_CONSTRAINT_COMPARISON_TOLERANCE = 0.00001;
+const double GlobalConfiguration::RELU_CONSTRAINT_COMPARISON_TOLERANCE = 1e-6;
```
AleksandarZeljic, line 73: This is a slightly smaller value. It might make sense to make this change in a separate PR so that is obvious, as it might affect performance (I think it's unlikely), but it's a cleaner approach in case it does affect performance.

AleksandarZeljic (approved):
LGTM. See the comment re: GlobalConfiguration values.

anwu1219:
> LGTM. See the comment re: GlobalConfiguration values.

I agree it's cleaner to do this in a different PR in case there's big performance change. Thanks for pointing this out! I reverted that particular change and will submit a separate PR.

PR #513, opened 2022-02-03
Add bound-aware auxiliary constraints after preprocessing 

anwu1219:
For Sign constraint, currently, we are not adding the symbolic lower and upper bound as auxiliary constraints. 
Add those constraints seem to have positive effects and also makes the SoI cost term for Sign much cleaner (just 1 - f for positive phase and 1 + f for negative phase).
 
This PR:
1. Extend PLConstraint with a addAuxEquationsAfterPreprocessing method().
2. Implement this method for Sign, which adds the symbolic upper and lower bounds in the BNN paper as constraints.
2. Reduce the dependency of NLR bound propagation passes on the Tableau. This is needed for this PR's purpose because we want to add tight symbolic upper/lower bounds, which means we want to add those aux equations to the preprocessedQuery after performing the SBT but before initializing the tableau.
3. Add 19 regression tests for BNN to regress1 using the MLP from the BNN paper. With the Aux constraints, those regression tests took ~1 minutes to run in total. Without the Aux constraints for Sign, two of those tests takes more than 5 minutes (time out set to 5 minutes).


AleksandarZeljic:
@guyam2 We have trouble tagging you again. Seeing if this will let tag you as the reviewer.

guyam2:
@AleksandarZeljic try now :)

guykatzz (approved):
Overall, LGTM
With this mechanism for adding post-PP bounds, it is natural to add the upper edge of the triangle for ReLU constraints (so far for y=relu(x) we only have y>=x and y>=0, but we can also add the upper-bound equation on y that uses the bounds on x). Could you please see how that affects performance? It can also be a separate PR.

anwu1219:
> Overall, LGTM
> With this mechanism for adding post-PP bounds, it is natural to add the upper edge of the triangle for ReLU constraints (so far for y=relu(x) we only have y>=x and y>=0, but we can also add the upper-bound equation on y that uses the bounds on x). Could you please see how that affects performance? It can also be a separate PR.

I think last time I checked, there were indeed some moderate improvement with this upper bound for ReLU. I can make a separate PR for that.

PR #514, opened 2022-02-04
Extend Sign and Abs with SoI and make DeepSoI default.

anwu1219:
The regression test for BNN would probably fail for now, because of the missing aux constraints. They should pass when PR https://github.com/NeuralNetworkVerification/Marabou/pull/513
is integrated.


guykatzz (approved):

src/common/LinearExpression.h
```diff
@@ -39,6 +39,8 @@ class LinearExpression
 
     double evaluate( const Map<unsigned, double> &assignment );
 
+    bool isZero();
```
guykatzz, line 42: Make this a const function

src/engine/Engine.cpp
```diff
@@ -320,6 +320,7 @@ bool Engine::solve( unsigned timeoutInSeconds )
             _tableau->toggleOptimization( false );
             continue;
         }
+        /*
```
guykatzz, line 323: Is this commenting-out on purpose?

src/engine/SumOfInfeasibilitiesManager.cpp
```diff
@@ -136,6 +143,23 @@ void SumOfInfeasibilitiesManager::initializePhasePatternWithCurrentInputAssignme
...
 
+void SumOfInfeasibilitiesManager::initializePhasePatternWithCurrentAssignment()
+{
+    obtainCurrentAssignment();
+
+    for ( const auto &plConstraint : _plConstraints )
+    {
+        ASSERT( !_currentPhasePattern.exists( plConstraint ) );
+        if ( plConstraint->isActive() && !plConstraint->phaseFixed() &&
+             plConstraint->getType() != DISJUNCTION )
```
guykatzz, line 154: I don't like the SoIManager probing for exact constraint types. Better have something like a virtual function plc->supportsSoi(), and have the SoIManager call it to check

src/system_tests/Test_Disjunction.h
```diff
@@ -134,8 +134,8 @@ class DisjunctionTestSuite : public CxxTest::TestSuite
         Engine engine;
         TS_ASSERT_THROWS_NOTHING( engine.processInputQuery( inputQuery ) );
 
-        TS_ASSERT_THROWS_NOTHING( engine.solve() );
-
+        engine.solve();
+        return;
```
guykatzz, line 138: A left over?

PR #515, opened 2022-02-04
make all disjuncts bounds

anwu1219:
Add a processing pass to Disjunction constraints so that the disjuncts only contain bounds. This not only makes it possible to do SoI on Disjunctions but also makes Disjunction compatible with the context dependent refactoring (no new vars introduced during case splits).


guykatzz (commented):
The functionality added here seems correct, but I dislike making the Preprocessor aware of specific constraints (Disjunctions, in this case). Why not place this functionality within the disjunction itself? Whenever it's initialized form a set of PLCaseSplits, just transform them as part of the c'tor. Alternatively, have a general PLC method called from the Preprocessor, allowing PLCs to transform themselves.

anwu1219:
> The functionality added here seems correct, but I dislike making the Preprocessor aware of specific constraints (Disjunctions, in this case). Why not place this functionality within the disjunction itself? Whenever it's initialized form a set of PLCaseSplits, just transform them as part of the c'tor. Alternatively, have a general PLC method called from the Preprocessor, allowing PLCs to transform themselves.

I see. That makes sense! I think I’ll go with the second solution, because we need to store the newly generated equations in the inputQeury.

src/engine/DisjunctionConstraint.h
```diff
@@ -144,6 +144,12 @@ class DisjunctionConstraint : public PiecewiseLinearConstraint
     */
     void getEntailedTightenings( List<Tightening> &tightenings ) const override;
 
+    /*
+      Transform the disjunction into a disjunction where each disjunct only
+      contains variable bounds.
+    */
+    void transformIfNeeded( InputQuery &inputQuery ) override;
```
AleksandarZeljic, line 151: The function name should reflect what the function does, IfNeeded does not tell the reader much, functions are in general invoked when needed and no addition information is conveyed when it is needed. The latter can be discovered by searching where the function is used. 

Perhaps something along the lines of rewriteDisjunctionUsingAux / introduceAuxForDisjuncts?

src/engine/DisjunctionConstraint.cpp
```diff
@@ -182,6 +183,85 @@ PiecewiseLinearCaseSplit DisjunctionConstraint::getValidCaseSplit() const
...
+        /*
+          Given a constraint AX >= b in the disjunct, we introduce an auxiliary
+          variable aux, add new linear constraints AX + aux = b and change the
+          constraint in the disjunct to aux <= 0
+
+          The LE case is symmetric.
+        */
+        for ( const auto &equation : equationsToProcess )
+        {
+            unsigned aux = inputQuery.getNumberOfVariables();
```
AleksandarZeljic, line 224: Lines 224-225 are probably a common pattern that could be a new function in InputQuery. Although this PR might not be the place to do it.

anwu1219: Yeah, these two lines are called in different places. Maybe worth doing a pass in a separate PR.

src/engine/DisjunctionConstraint.h
anwu1219: What you suggested is what I had, but Guy's suggested that we "have a general PLC method called from the Preprocessor, allowing PLCs to transform themselves.". I think the documentation should make it clear what this method does in the DisjunctionConstraint?

src/engine/tests/Test_DisjunctionConstraint.h
```diff
@@ -500,4 +501,141 @@ class DisjunctionConstraintTestSuite : public CxxTest::TestSuite
...
+        List<PiecewiseLinearCaseSplit> splitsAfter = disj->getCaseSplits();
+
+        TS_ASSERT_EQUALS( casesBefore.size(), casesAfter.size() );
+        TS_ASSERT_EQUALS( splitsBefore.size(), splitsAfter.size() );
+
+        // 1 <= x0 <= 5, x1 = x0
+        // 5 <= x0 , x1 = 2x2 + 5
+        // x1 - x0 <= 1, x1 + x2 >= 2
+
+        // In total there are 4 (in)equalities in all the disjuncts.
```
AleksandarZeljic, line 544: This comment should be clarified or moved, depending on its intent. It is confusing given the assertion two lines below.

src/engine/tests/Test_DisjunctionConstraint.h
```diff
@@ -500,4 +501,141 @@ class DisjunctionConstraintTestSuite : public CxxTest::TestSuite
...
+        TS_ASSERT_EQUALS( inputQuery.getNumberOfVariables(), 9u );
+        TS_ASSERT_EQUALS( inputQuery.getEquations().size(), 6u );
```
AleksandarZeljic, line 546: Should these assertions have their before version, to highlight what is the difference (if any) during the transformation?

AleksandarZeljic (approved):
LGTM

src/engine/tests/Test_DisjunctionConstraint.h
anwu1219: Will add the before assertions.

src/engine/tests/Test_DisjunctionConstraint.h
anwu1219: Will clarify the documentation.

src/engine/tests/Test_DisjunctionConstraint.h
anwu1219: done.

src/engine/tests/Test_DisjunctionConstraint.h
anwu1219: done.

src/engine/DisjunctionConstraint.h
anwu1219: changed to rewriteToUseAuxIfNeeded()

PR #519, opened 2022-02-11
Revive max constraint

anwu1219:
My attempt to clean up the MaxConstraint class.
Unit tests pretty much remains unchanged except that 
1) we are testing MaxConstraints' behaviors after "transformToUseAuxVariablesIfNeeded()".
2) there are a few assertions in the original unit test which I think are not correct.

I think in the next few PRs about context dependent object, we should probably just change addAuxilliraryEquations() to transformToUseAuxVariablesIfNeeded() in every constraint class. Because not using aux variables is no longer an option with the context dependent object.

 Invariants to maintain in this class (also written in the header file):
 1. All methods called after transformToUseAuxVariablesIfNeeded are operating under the assumption that f >= element for each element in _element, and f >= _maxValueofEliminatedVariables
 2. _maxLowerBound keeps track of the maximal lower bound of the output of  the MaxConstraint, it is updated-on-the-fly by notifyLowerBound(), notifyUpperBound(), eliminateVariable()
 3. _phaseStatus is updated on-the-fly by notifyLowerBound(), notifyUpperBound(), eliminateVariable().
 4. elements in _elements are feasible (wrt. current variable bounds) Both _elements and _haveFeasibleEliminatedVariables are updated on-the-fly in notifyLowerBound(), notifyUpperBound(), eliminateVariable().
 5. The constraint is _obsolete only when 1) all elements are eliminated (handled by eliminateVariable()); 2) _f is eliminated.

Please also note that the case where the output variable is part of the input f = max (f, x1, x2...) is handled by transformtoUseAuxVariablesIfNeeded(). The method will add the constraint f >= x1, f>=x2 and make the constraint obsolete.


src/engine/MaxConstraint.cpp
```diff
@@ -837,16 +665,35 @@ String MaxConstraint::serializeToString() const
...
+        if ( _elementToAux.exists( variable ) )
+        {
+            unsigned aux = _elementToAux[variable];
+            _elementToAux.erase( variable );
+            _auxToElement.erase( aux );
+        }
+    }
+}
+
+
```
guykatzz, line 696: remove empty line

src/engine/MaxConstraint.cpp
```diff
@@ -837,16 +665,35 @@ String MaxConstraint::serializeToString() const
...
 bool MaxConstraint::haveOutOfBoundVariables() const
 {
     double fValue = _assignment.get( _f );
```
guykatzz, line 699: Need to address aux variables here, too?

guykatzz (commented):
I left a bunch of comments - most are cosmetic, but some are real.
One question that I had that didn't make it into the comments:

in MaxConstraint::getPhaseStatusInAssignment, if f currently has an assignment greater than all the elements, we return phase = ELIMINATED, even if there are no eliminated variables. Is this intended? If it is, maybe add a comment.

src/engine/Engine.cpp
```diff
@@ -2608,11 +2609,11 @@ bool Engine::solveWithMILPEncoding( unsigned timeoutInSeconds )
             applyAllBoundTightenings();
             applyAllValidConstraintCaseSplits();
         }
-        do
+
+        while ( applyAllValidConstraintCaseSplits() )
```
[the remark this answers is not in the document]
anwu1219: The processInputQuery performed SBT and we haven't called applyAllValidConstraintCaseSplits() yet

src/engine/Engine.cpp
```diff
@@ -2101,7 +2101,8 @@ List<unsigned> Engine::getInputVariables() const
 
 void Engine::performSimulation()
 {
-    if ( _simulationSize == 0 || !_networkLevelReasoner )
+    if (  _simulationSize == 0 || !_networkLevelReasoner ||
```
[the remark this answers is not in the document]
anwu1219: done

src/engine/MaxConstraint.cpp
```diff
@@ -406,8 +374,14 @@ bool MaxConstraint::satisfied() const
     if ( !( _assignment.exists( _f ) && _assignment.size() > 0 ) )
```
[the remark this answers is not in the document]
anwu1219: I guess we really want every assignment to be present because I'm not sure we can safely conclude the constraint is satisfied otherwise. Updated to check this.

src/engine/MaxConstraint.cpp
```diff
@@ -406,8 +374,14 @@ bool MaxConstraint::satisfied() const
...
         throw MarabouError( MarabouError::PARTICIPATING_VARIABLES_ABSENT );
 
+    auto byAssignment = [&](const unsigned& a, const unsigned& b) {
```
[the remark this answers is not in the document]
anwu1219: done

src/engine/MaxConstraint.cpp
```diff
@@ -406,8 +374,14 @@ bool MaxConstraint::satisfied() const
...
+                            return _assignment[a] < _assignment[b];
+                        };
+    double maxValue =  _assignment.get( *std::max_element( _elements.begin(),
```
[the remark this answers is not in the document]
anwu1219: done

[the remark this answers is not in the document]
anwu1219: I think this line is getting the *element* with the maximal assignment?

src/engine/MaxConstraint.cpp
anwu1219: done

src/engine/MaxConstraint.h
```diff
@@ -262,10 +246,28 @@ class MaxConstraint : public PiecewiseLinearConstraint
                  : static_cast<unsigned>( phase ) - MAX_VARIABLE_TO_PHASE_OFFSET;
```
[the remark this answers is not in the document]
anwu1219: I think this was introduced in the last refactoring by Aleks. Since Phase_not_fixed has value "zero", with the offset x0 will have phase status "1".

src/engine/MaxConstraint.cpp
anwu1219: Done.

src/engine/MaxConstraint.cpp
```diff
@@ -121,64 +123,83 @@ void MaxConstraint::registerAsWatcher( ITableau *tableau )
...
+                                 _maxValueOfEliminatedPhases );
         }
     }
 
-    if ( maxErased )
-        resetMaxIndex();
+    if ( phaseFixed() )
+        _phaseStatus = ( _haveFeasibleEliminatedPhases ?
+                         MAX_PHASE_ELIMINATED :
+                         variableToPhase( *_elements.begin() ) );
```
AleksandarZeljic, line 198: Do you need brackets around the ternary expression?

src/engine/MaxConstraint.cpp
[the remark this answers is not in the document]
guykatzz: Still not convinced. std::max_element doesn't distinguish between max's f and its elements, no?
I think if you have f=max(x1,x2), and currently f=7, x1=x2=1, you would get maxValue = 7.

guykatzz:
LGTM, except for my one pending comment, and Aleks' comment.

src/engine/MaxConstraint.cpp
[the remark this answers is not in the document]
anwu1219: I still think the code is correct, since "_elements" contains only input variables to Max and does not include _f. But I do think the code is more confusing than it needs to be. I got rid of the std::max_element business.

src/engine/MaxConstraint.h
```diff
@@ -160,10 +175,9 @@ class MaxConstraint : public PiecewiseLinearConstraint
     void getEntailedTightenings( List<Tightening> &tightenings ) const override;
 
     /*
-      For preprocessing: get any auxiliary equations that this constraint would
-      like to add to the equation pool.
+      Represent the Max constraint using the aux variables.
     */
-    void addAuxiliaryEquations( InputQuery &inputQuery ) override;
+    void transformToUseAuxVariablesIfNeeded( InputQuery &inputQuery ) override;
```
AleksandarZeljic, line 180: Update function name per the offline discussion.

src/engine/MaxConstraint.h
```diff
@@ -190,54 +204,24 @@ class MaxConstraint : public PiecewiseLinearConstraint
...
-    inline void clearMaxIndex()
-    {
-        setPhaseStatus( PHASE_NOT_FIXED );
-    }
+    // We eagerly keep track of the max lower bound of the inputs.
+    double _maxLowerBound;
 
-    void resetMaxIndex();
+    // We keep track of eliminated variables and the maximal value.
+    bool _haveFeasibleEliminatedPhases;
```
AleksandarZeljic, line 223: Adjectives feasible and eliminated might sounds a bit contradictory. Can we come up with a simpler name or just elaborate on the exact meaning in the comments.

src/engine/MaxConstraint.h
```diff
@@ -190,54 +204,24 @@ class MaxConstraint : public PiecewiseLinearConstraint
...
 
-    inline void clearMaxIndex()
-    {
-        setPhaseStatus( PHASE_NOT_FIXED );
-    }
+    // We eagerly keep track of the max lower bound of the inputs.
+    double _maxLowerBound;
 
-    void resetMaxIndex();
+    // We keep track of eliminated variables and the maximal value.
```
AleksandarZeljic, line 222: Use consistent comment syntax.

src/engine/MaxConstraint.h
```diff
@@ -262,10 +246,28 @@ class MaxConstraint : public PiecewiseLinearConstraint
...
     /*
       Return true iff f or the elements are not all within bounds.
     */
     bool haveOutOfBoundVariables() const;
+
+    /*
+     * Returns a boolean value indicating if at least one input variable was eliminated (True)
+     * or not (False)
+     */
+    bool wereVariablesEliminated() const;
```
AleksandarZeljic, line 263: Maybe hasEliminatedVariables()? In terms of OO design, the question wereVariablesEliminated() implicitly involves time, which is unnecessary.

src/engine/MaxConstraint.h
```diff
@@ -262,10 +246,28 @@ class MaxConstraint : public PiecewiseLinearConstraint
...
+
+    /*
+      For DEBUG only:
+      Return true if all the element in _elements are feasible
+      based on their bounds.
+    */
+    void checkAllInputElementsFeasible() const;
```
AleksandarZeljic, line 270: Since its used only for debugging, maybe use assert rather then check to make that distinction? 

Is AllInputElements in the function name redundant? While the max variable might be an element, by itself it can never be infeasible, so making that distinction may be unnecessary. What do you think?

src/engine/MaxConstraint.cpp
```diff
@@ -121,64 +123,83 @@ void MaxConstraint::registerAsWatcher( ITableau *tableau )
...
+    */
+    if ( _auxToElement.exists( variable ) )
     {
-        _maxLowerBound = value;
-        List<unsigned> toRemove;
-        for ( auto element : _elements )
+        // The case that this variable is an aux variable.
+        // We can eliminate the corresponding case if the aux variable is
+        // positive.
+        if ( FloatUtils::isPositive( value ) )
```
AleksandarZeljic, line 161: The ifs on line 156 and 162 can merged. Comment on line 158 seems redundant.

Re: Comment style - we should probably use /*  */  for function descriptions (here used on line 153-155), and use // for line comments in the actual implementation. The extra space used by /**/ is useful in header for the reader to distinguish (groups of) methods visually, while the same amount of space within code itself would be detracting from easy readability.

src/engine/MaxConstraint.cpp
```diff
@@ -121,64 +123,83 @@ void MaxConstraint::registerAsWatcher( ITableau *tableau )
...
+            eliminateCase( _auxToElement[variable] );
+    }
+    else if ( variable == _f || _elements.exists( variable ) )
+    {
+        // If the variable is either in _element or _f. The only case that this is
+        // going to eliminate case is that the new lowerBound is greater than the
+        // _maxLowerBound.
+        // We update _maxLowerBound, and go through each element to see if it
+        // can be eliminated.
+        // We also update the _haveFeasibleEliminatedCase flag.
```
AleksandarZeljic, line 171: This comment can be streamlined:

// When the variable is either in _element or _f, a case is eliminated,
//  when the new lowerBound is greater than the _maxLowerBound.

The comment should not be stating things that are obvious when reading the code. Lines 169-170 would be somewhat appropriate if this was a method description, but here the code follows and the reader can see what is done.

src/engine/MaxConstraint.cpp
```diff
@@ -121,64 +123,83 @@ void MaxConstraint::registerAsWatcher( ITableau *tableau )
...
-    if ( maxErased )
-        resetMaxIndex();
+    if ( phaseFixed() )
+        _phaseStatus = ( _haveFeasibleEliminatedPhases ?
+                         MAX_PHASE_ELIMINATED :
+                         variableToPhase( *_elements.begin() ) );
 
+    /*
+      Now do some bound tightening.
+    */
```
AleksandarZeljic, line 202: Re consistency: To clarify my previous comment, I think this is an OK use of the /* */ comments.

src/engine/MaxConstraint.cpp
```diff
@@ -403,204 +370,71 @@ unsigned MaxConstraint::getF() const
...
-        throw MarabouError( MarabouError::PARTICIPATING_VARIABLES_ABSENT );
+    DEBUG({
+            if ( !( _assignment.exists( _f ) ) )
+                throw MarabouError
+                    ( MarabouError::PARTICIPATING_VARIABLES_ABSENT,
+                      Stringf( "f(x%u) assignment missing.", _f ).ascii() );
+            for ( const auto &element : _elements )
+                if ( !( _assignment.exists( element ) ) )
+                    throw MarabouError
+                        ( MarabouError::PARTICIPATING_VARIABLES_ABSENT,
```
AleksandarZeljic, line 381: In the Error name, absent implies the variable is missing, while in reality assignment is missing. So maybe replace ABSENT with MISSING_ASSIGNMENT?

src/engine/MaxConstraint.h
anwu1219: Done

src/engine/MaxConstraint.h
anwu1219: Done

src/engine/MaxConstraint.h
anwu1219: Added more documentation. I thought more about "_isEliminatedPhaseFeasible" but I didn't like it too much either because it doesn't clearly signal that the flag is on iff 1) there are eliminated phases; and 2) one of them is feasible.

src/engine/MaxConstraint.h
anwu1219: Removed this method since it's no longer used anywhere.

src/engine/MaxConstraint.h
anwu1219: Removed this method since I didn't end up using it.

src/engine/MaxConstraint.cpp
anwu1219: Done.

src/engine/MaxConstraint.cpp
anwu1219: Done

src/engine/MaxConstraint.cpp
anwu1219: Renamed to participating_var_missing_assignment

PR #520, opened 2022-02-11
Add BoundManager object to the Tableau

AleksandarZeljic:
This PR passed a BoundManager object to the Tableau which is not used by default. 
The PR introduces  _useBoundManager flag and Tableau::enableBoundManager() 
method (for purposes of testing and integration).


anwu1219 (approved):
LGTM.

PR #524, opened 2022-02-16
Add option to solve LP with Gurobi

anwu1219:
This PR add option to solve the LPs with Gurobi during the search. This is controlled by a flag --lp-solver={native/gurobi}
When Gurobi is enabled, LP is solved by Gurobi by default When Gurobi is not enabled, LP is solved by native simplex engine.

1. When Gurobi is used for LP, Tableau only functions as a centralized place to store bounds.
2. In PLconstraint, access of _assignment are abstracted with "getAssignment()" methods.


guykatzz (approved):
There's a stale exception comment in MILPEncoder that says that only Relus and Maxes are supported - maybe just say "unsupported constraint type", instead.

PR #526, opened 2022-02-16
More Github Action

anwu1219:
Add a production build to CI. Sometimes variables that are used in Debug mode might not be used in Production mode, which result in compilation error in Production build.


AleksandarZeljic (approved):
LGTM.

PR #532, opened 2022-02-20
Do not copy tableau state during push and pop

anwu1219:
- Add assertions in SmtCore that all case splits applying during push and pop contain only bound updates.
- Due to this assertion, copying of the tableau state is no longer necessary. In particular, when backtracking, we just need to restore the variable bounds from the previous level and the other tableau state do not need to change. 
- Copying of the entire tableau state is used only for precision restoration. 
- Remove the VariableWatcher::notifyVariableValue() function, but instead get the variable assignment directly from the Tableau.


src/engine/Tableau.cpp
```diff
@@ -1729,27 +1738,41 @@ void Tableau::storeState( TableauState &state ) const
...
 
-        _boundsValid = true;
+        if ( _lpSolverType == LPSolverType::NATIVE )
+        {
+            // The bounds might be invalid in the state from which we backtrack,
+            // This means we might need to fix the non-basic variable assignments
+            for ( unsigned i = 0; i < _n - _m; ++i )
+            {
+                unsigned variable = _nonBasicIndexToVariable[i];
+                updateVariableToComplyWithLowerBoundUpdate( variable,
```
anwu1219, line 1767: @guykatzz If I don't do this update, the Tableau invariant can be broken (some non-basic variable will go out of bound). Are there better solution than this? 
The alternative would be to not call updateVariableToComplyWithLowerBoundUpdate() when _boundsValid is false in the "tightenLower/UpperBounds methods". But doing that seems to result in invariant being broken elsewhere.
I also tried throw InfeasibleQueryException() in those tightenbounds method once the bounds become invalid. But that is also problematic, because now this exception could be thrown in "popSplit()" method..

src/engine/DnCWorker.cpp
```diff
@@ -49,7 +50,8 @@ DnCWorker::DnCWorker( WorkerQueue *workload, std::shared_ptr<IEngine> engine,
 
     // Obtain the current state of the engine
     _initialState = std::make_shared<EngineState>();
-    _engine->storeState( *_initialState, true );
+    _engine->storeState( *_initialState,
+                         TableauStateStorageLevel::STORE_ALL_TABLEAU_STATE );
```
guykatzz, line 54: all -> entire?

src/engine/Engine.cpp
```diff
@@ -1364,6 +1363,12 @@ bool Engine::processInputQuery( InputQuery &inputQuery, bool preprocess )
             }
         }
 
+        for ( const auto &constraint : _plConstraints )
+        {
+            constraint->registerTableau( _tableau );
+        }
+
```
guykatzz, line 1370: remove empty line

src/engine/PiecewiseLinearConstraint.h
```diff
@@ -582,16 +588,19 @@ class PiecewiseLinearConstraint : public ITableau::VariableWatcher
     /**********************************************************************/
     inline bool existsAssignment( unsigned variable ) const
     {
-        return ( _gurobi == nullptr ) ?
-            _assignment.exists( variable )
-            : _gurobi->existsAssignment( Stringf( "x%u", variable ) );
+        if ( _gurobi == nullptr && _tableau != nullptr )
```
guykatzz, line 591: This is intended to give preference to Gurobi, right? So lets reorder:

if ( gurobi exists )
   return gurobi value;

if ( tableu exists )
   return tableu value

fail

src/engine/Tableau.cpp
guykatzz: I guess an alternative would be to store the variables' assignment as part of the stored state? And then have another invariant that says that a stored state is valid, as far as the NB assignments are concerned?

But why do you want to change this? I don't think it's that expensive, performance wise

guykatzz (approved):
See few comments inside

src/engine/Tableau.cpp
anwu1219: Ah ok. I think I'll keep the current solution.

src/engine/DnCWorker.cpp
anwu1219: done

src/engine/Engine.cpp
anwu1219: done

src/engine/PiecewiseLinearConstraint.h
anwu1219: Done. Yeah this reads much better..

PR #534, opened 2022-02-23
bug fix in Max

anwu1219:


src/engine/MaxConstraint.cpp
```diff
@@ -231,10 +231,10 @@ void MaxConstraint::notifyUpperBound( unsigned variable, double value )
         if ( FloatUtils::isZero( value ) )
         {
             unsigned currentElement = _auxToElement[variable];
-            for ( const auto &element : _elements )
-                if ( element != currentElement )
-                    eliminateCase( element );
```
anwu1219, line 236: eliminateCase removes the element from _elements, which is problematic..

AleksandarZeljic (approved):
LGTM

PR #539, opened 2022-03-02
Group the options, updateReadMe

anwu1219:
Some house-keeping..
1. Group the options into _positional, _common, _other, _expert. The --help message looks like the following now:
Screenshot from 2022-03-02 10-49-25
2. Update the Gurobi version.
3. Add information about Marabou configurations in the ReadMe


src/configuration/OptionParser.h
```diff
@@ -57,8 +57,13 @@ class OptionParser
 private:
     boost::program_options::variables_map _variableMap;
     boost::program_options::options_description _optionDescription;
+    boost::program_options::options_description _positional;
+    boost::program_options::options_description _common;
+    boost::program_options::options_description _other;
+    boost::program_options::options_description _expert;
     boost::program_options::positional_options_description _positionalOptions;
 
+
```
guykatzz, line 66: remove empty line

guykatzz:
Hi Andrew,
This looks good to me - thanks for taking care of this.
Notice that the tests are failing.

I would add the default value for all parameters (currently it's only stated for some).
In the long list, maybe add empty lines to group together related options - SOI options, SNC options, etc.

Do we need the preprocessor bound tolerance parameter? I've never changed it; have you?

Some options are conditions upon others - e.g., reluplex split threshold and the splitting strategy. This is worth mentioning.

Cosmetics: lets have a period '.' at the end of every line, or none at all, for uniformity.

guykatzz (approved):

anwu1219:
>  Notice that the tests are failing.

Yeah I'm taking a look now!

> I would add the default value for all parameters (currently it's only stated for some). In the long list, maybe add empty lines to group together related options - SOI options, SNC options, etc.

These print outs are automatically generated by Boost.. and I think the default values won't be shown if the option is a boolean flag or the default value is empty (""). I can further break down the expert options into more options_description objects.

> Do we need the preprocessor bound tolerance parameter? I've never changed it; have you?

Yeah, I'm not sure whether we need to expose this. I've never changed its value.

> Some options are conditions upon others - e.g., reluplex split threshold and the splitting strategy. This is worth mentioning.

I'll update the documentation to mention this.

> Cosmetics: lets have a period '.' at the end of every line, or none at all, for uniformity.

Will do.

src/configuration/OptionParser.h
anwu1219: done

PR #542, opened 2022-03-03
Use context-dependent BoundManager to store bounds in Tableau

AleksandarZeljic:
This PR transitions Marabou to the first stage of using Context to synchronize the search. Context is added to the SmtCore which controls the synchronization, enabling the Tableau to use context dependent BoundManager to store the bounds and removing the memory management overhead. 

The PR includes the following changes:
- Introduce IBoundManager and MockBoundManager.
- Adds local bounds to BoundManager for expediency of access by Tableau.
- Use BoundManager to store bounds in Tableau.
- Rename and move Tableau::_validBounds to BoundManager::_consistentBounds.
- BoundManager::set*Bounds methods no longer throw InfeasibleQueryException and set the _consistentBounds flag instead.
- Add Context to SmtCore and synchronize with the search correctly using postContextHook to maintain Tableau invariants.
-  SmtCore continues backtracking immediately if implied CaseSplit results in inconsistent bounds.
- Remove local bounds in the Tableau.
- Initialize BoundManager correctly.


src/engine/BoundManager.cpp
```diff
@@ -27,7 +27,10 @@ BoundManager::BoundManager( Context &context )
     : _context( context )
     , _size( 0 )
     , _tableau( nullptr )
+    , _consistentBounds( &_context )
+    , _firstInconsistentTightening( 0, 0.0, Tightening::LB )
 {
+  _consistentBounds = true;
```
anwu1219, line 33: Use 4-space instead for 2-space for indentation to be consistent with rest of the code base?

src/engine/BoundManager.cpp
```diff
@@ -94,6 +97,15 @@ bool BoundManager::tightenUpperBound( unsigned variable, double value )
     return tightened;
 }
 
+void BoundManager::recordInconsistentBounds( unsigned variable, double value, Tightening::BoundType type )
```
anwu1219, line 100: Bounds=> Bound ? 
Since we are only recording one bound here.

src/engine/Engine.h
```diff
@@ -719,6 +719,13 @@ class Engine : public IEngine, public SignalHandler::Signalable
     */
     bool minimizeCostWithGurobi( const LinearExpression &costFunction );
 
+    /*
+      Get Context reference
+    */
+    Context &getContext() { return _context; }
+
+    bool consistentBounds() const;
```
anwu1219, line 727: This method doesn't seem to be used anywhere.

src/engine/Tableau.h
```diff
@@ -118,7 +113,6 @@ class Tableau : public ITableau, public IBasisFactorization::BasisColumnOracle
     unsigned getN() const;
 
     /*
-      Check if an assignment exists for the variable.
```
anwu1219, line 121: Was this line intentionally removed?

src/engine/BoundManager.cpp
AleksandarZeljic: Done.

src/engine/Engine.h
AleksandarZeljic: Both methods are used in the SmtCore, getContext() in the ctor and consistentBounds() on line 287.

src/engine/Tableau.h
AleksandarZeljic: Oops, n - restored.

anwu1219 (approved):
LGTM

PR #543, opened 2022-03-03
Put back random seed setting. 

anwu1219:
Accidentally removed random seed setting in single-threaded mode.


AleksandarZeljic (approved):
LGTM

PR #548, opened 2022-03-08
Reduce memory usage

anwu1219:
Reduce the number of input query and tableau state copy.

### DIFF
--- a/src/configuration/Options.cpp
+++ b/src/configuration/Options.cpp
@@ -48,7 +48,7 @@ void Options::initializeDefaultValues()
     _boolOptions[DUMP_BOUNDS] = false;
     _boolOptions[SOLVE_WITH_MILP] = false;
     _boolOptions[PERFORM_LP_TIGHTENING_AFTER_SPLIT] = false;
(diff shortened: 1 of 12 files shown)


src/configuration/Options.cpp
```diff
@@ -48,7 +48,7 @@ void Options::initializeDefaultValues()
     _boolOptions[DUMP_BOUNDS] = false;
     _boolOptions[SOLVE_WITH_MILP] = false;
     _boolOptions[PERFORM_LP_TIGHTENING_AFTER_SPLIT] = false;
-    _boolOptions[PARALLEL_DEEPSOI] = true;
```
guykatzz, line 51: I think "positive" flags are easier to understand than negative ones, due to double negations. Why was it bothering you the way it was before?

guykatzz (approved):

src/configuration/Options.cpp
anwu1219: Somehow boost option only seems to be able to set the flag to true. I played with it but didn't get it to work the other way..

PR #549, opened 2022-03-08
Set bounds in Tableau after registering watchers and dump bounds after tableau is registered in NLR

anwu1219:
This will allow constraint bound tightening to work from the very beginning.


tagomaru (approved):
LGTM

PR #553, opened 2022-03-29
Add command line ONNX support

### LINKED ISSUE #498: Advice on adding support for `.onnx` files to the command line interface
I have multiple onnx networks I would like to verify, and I cannot use MarabouPy as I am not working in a Python environment. One approach would be to convert them to the [NNet format](https://github.com/sisl/NNet) but that repo seems to be very behind the times (e.g. uses Tensorflow 1) and no matter what I do I can't get it to work, and anyway it's in Python so again I can't include it as a a regular part of my workflow.

So I am about to embark on a possibly fool-hardy quest to add support for onnx files to the command line interface for Marabou. My plan is as follows:

- Add a new `OnnxParser.h/cpp` to `src/input_parsers`
- Add a new conditional at the point below to use either the `AcasParser` (I guess should really be called `NNetParser`) or the `OnnxParser` depending on the extension of the network file. https://github.com/NeuralNetworkVerification/Marabou/blob/ffd353b09ff1e36deca1ca51cd2d93fbf5c97c71/src/engine/Marabou.cpp#L91-L93

 I was wondering if anyone had any advice or whether there were any objections to adding `onnx` as a dependency to Marabou?

MatthewDaggitt:
Fixes #498.

Pretty much a straight a port as `MarabouNetworkONNX.py` as I could manage. Obviously Python's dynamic type system and numpy's support for array wrangling makes porting it one for one difficult. In the end I stored everything as flat 1D arrays and then indexed into them as if they were multi-dimensional. 

Immediate outstanding issues:
 - Haven't yet added in support for `maxPool`, `conv` and `gemm` layers. Thought it better to get the basics merged in and then come back and add support later.
 - Although I've tested it on the examples I want to use it for, I haven't managed to plug it into Marabou's test suite. Any advice on how best to do that?
 - I've followed `MarabouNetworkONNX.py`'s lead in ignoring a lot of optional parameters specified in the [ONNX docs](https://github.com/onnx/onnx/blob/main/docs/Operators.md). It's really not great as it will silently ignore them, but it's no worse then what the existing Python implementation does.

Just a heads-up I would appreciate a thorough review of this! As previously mentioned, first time working in C++, think I've got my head around it but no guarantees I'm not doing something... _interesting_!


MatthewDaggitt:
@AleksandarZeljic / @guykatzz is there anything I can do to help get the ball rolling on this getting merged in?

AleksandarZeljic:
Hi @MatthewDaggitt, thanks for bumping this! I can start reviewing the PR later this week and then we can discuss how to proceed.

src/engine/Marabou.h
```diff
@@ -56,6 +57,11 @@ class Marabou
     */
     AcasParser *_acasParser;
 
+    /*
+      ACAS network parser
```
AleksandarZeljic, line 61: Update comment to Onnx parser.

src/input_parsers/NetworkParser.cpp
```diff
@@ -0,0 +1,166 @@
...
+    // TODO check this last two
+    for (std::pair<Variable,float> lower : _lowerBounds)
+    {
+        assert(lower.first < _numVars);
+        query.setLowerBound(lower.first,lower.second);
+    }
+
+    for (std::pair<Variable,float> upper : _upperBounds)
+    {
+        assert(upper.first < _numVars);
```
AleksandarZeljic, line 147: Marabou uses its own ASSERT macro from Debug.h instead of assert().

.gitignore
```diff
@@ -19,10 +19,13 @@ tools/boost_1_68_0
 /tools/boost.unzipped
 /tools/boost_1_68_0.tar.gz
 /tools/pybind11-2.3.0
+/tools/protobuf*
 /tools/OpenBLAS-0.3.9/
 /tools/OpenBLASv0.3.9.tar.gz
+/tools/onnx-1.10.2
 /cpp_interface_example/example.elf
 /src/input_parsers/mps_example/mps.elf
+.vscode
```
AleksandarZeljic, line 28: This line belongs in your local .gitignore.

deps/CVC4/base/check.h
```diff
@@ -206,7 +206,7 @@ class AssertArgumentException : public Exception
 
 #define Unimplemented() CVC4_FATAL() << "Unimplemented code encountered"
 
-#define InternalError() CVC4_FATAL() << "Internal error detected"
+//#define InternalError() CVC4_FATAL() << "Internal error detected"
```
AleksandarZeljic, line 209: Can you add a one line comment (both here and in the output.h) just saying that it was excluded because of collisions when using ... compiler/setup?

src/engine/Marabou.cpp
```diff
@@ -80,16 +89,27 @@ void Marabou::prepareInputQuery()
...
         {
             printf( "Error: the specified network file (%s) doesn't exist!\n", networkFilePath.ascii() );
             throw MarabouError( MarabouError::FILE_DOESNT_EXIST, networkFilePath.ascii() );
         }
         printf( "Network: %s\n", networkFilePath.ascii() );
 
-        // For now, assume the network is given in ACAS format
-        _acasParser = new AcasParser( networkFilePath );
-        _acasParser->generateQuery( _inputQuery );
+        if ( networkFilePath.endsWith( ".onnx" ) )
```
AleksandarZeljic, line 100: Maybe just cast networkFilePath to MString? You already implemented the method there and you can then get rid of the boost include, since it seems it is there solely for this call?

src/input_parsers/NetworkParser.h
```diff
@@ -0,0 +1,86 @@
+/*
```
AleksandarZeljic, line 1: Update the header to match template like in other files and provide a class description.

src/input_parsers/OnnxParser.h
```diff
@@ -0,0 +1,92 @@
...
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Guy Katz
+ ** This file is part of the Marabou project.
+ ** Copyright (c) 2017-2019 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved. See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** [[ Add lengthier description here ]]
```
AleksandarZeljic, line 12: Add a class description, perhaps provide a link to the ONNX/VNN-LIB format.

AleksandarZeljic (changes requested):
I've gone over all the files except for the two .cpp files which I will take a detailed look tomorrow. Overall, the PR looks pretty good. There are mostly cosmetic fixes to be made. 

One piece that should be included to finalize the PR is tests. You can take a look at /maraboupy/test/test_onnx.py and even re-purpose the tests from there.

Thanks for all the work so far and I appreciate your patience for the review.

MatthewDaggitt:
Thanks @AleksandarZeljic! I've implemented your suggestions so far. I'll have a look at plugging into the existing test suite today. Any chance you have the remainder of the review as well so we can wrap this up? :smile:

MatthewDaggitt:
Hmm after merging with the latest version of Marabou I'm now getting a fresh error internally in the boost libraries when trying to build the `Marabou` executable:
```
[  0%] Building CXX object CMakeFiles/MarabouHelper.dir/src/nlr/DeepPolyAnalysis.cpp.o
In file included from ./Marabou/tools/boost_1_68_0/installed/include/boost/thread/thread_only.hpp:17,
                 from ./Marabou/tools/boost_1_68_0/installed/include/boost/thread/thread.hpp:12,
                 from ./Marabou/tools/boost_1_68_0/installed/include/boost/thread.hpp:13,
                 from ./Marabou/src/nlr/DeepPolyAnalysis.cpp:33:
./Marabou/tools/boost_1_68_0/installed/include/boost/thread/pthread/thread_data.hpp: In member function ‘void boost::thread_attributes::set_stack_size(std::size_t)’:
./Marabou/tools/boost_1_68_0/installed/include/boost/thread/pthread/thread_data.hpp:60:19: error: comparison of integer expressions of different signedness: ‘std::size_t’ {aka ‘long unsigned int’} and ‘long int’ [-Werror=sign-compare]
   60 |           if (size<PTHREAD_STACK_MIN) size=PTHREAD_STACK_MIN;
      |                   ^
cc1plus: all warnings being treated as errors
```
I don't think that "all warnings being treated as errors" is a good idea to have turned on for _dependencies_ right? Have I misconfigured something?

AleksandarZeljic:
Hi Matthew, I am taking a look at what's going on with dependencies and will also provide the remainder of the review.

AleksandarZeljic:
I am still trying to figure out the issue with Boost -werror you encountered.

However, when I tried to compile Marabou the build time went up considerably by compiling the entirety of the protobuf library. Is there a way to only build essentials for the onnx parser? As it it the build time may be prohibitively long, especially in the context of CI.

MatthewDaggitt:
> I am still trying to figure out the issue with Boost -werror you encountered.

Thanks :heart: 

> However, when I tried to compile Marabou the build time went up considerably by compiling the entirety of the protobuf library. Is there a way to only build essentials for the onnx parser? As it it the build time may be prohibitively long, especially in the context of CI.

Yes, it does take a significant amount of time. In the latest commit, I've managed to speed it up by ~70% by using the magic `--enable-fast-install` option and disabling running the tests. However it's still takes ~4 minutes on my reasonably beefy laptop.

If that is insufficient, then a couple of suggestions?
 1. we could cache the result of the build in the CI github workflow?
 2. we could ask the user to install the pre-existing `protobuf-compiler` via `apt-get`?

AleksandarZeljic:
Regarding the -Werror issue - we use the download_boost.sh script in the tools folder to install boost and it does not inherit any flags from the CMakeLists.txt. Can you run that script in the tools folder and see if the issue persists? If it goes through without issues then: a) we indeed pass the flag somehow (which will be for me to figure out) and b) you will have boost installed and Marabou should not complain from that point onward. If you encounter the same -Werror situation then it would probably be a local issue.

I think that is good enough for now, we can revisit the question before we wrap up the PR and after I have consulted with the rest of the team. 

On a different note, I noticed that  you merged master into the branch. It's a good practice to rebase the working branch onto the master (using git rebase). Rebase replays the commits of one branch on top of the other, and by doing so ensures a linear history (i.e. without parallel commits). This is useful for history readability, but also for discovering when a bug is introduced using git bisect.

MatthewDaggitt:
> Can you run that script in the tools folder and see if the issue persists? If it goes through without issues then: a) we indeed pass the flag somehow (which will be for me to figure out) and b) you will have boost installed and Marabou should not complain from that point onward.

So I ran `download-boost.sh` manually. That ran with no errors. I then went back to the `build` folder and ran `cmake --build .` as before, but still got the same error. 

After a bit of fiddling I found that the place that is enabling the `werror` warning is the snippet below in the top-level `CMakeLists` file:
```
if (NOT MSVC)
    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
       set(COMPILE_FLAGS  -Wall -Wextra  -Werror -MMD -Qunused-arguments ) #-Wno-deprecated 
     else()
       set(COMPILE_FLAGS  -Wall -Wextra  -Werror -MMD ) #-Wno-deprecated 
    endif()
    set(RELEASE_FLAGS ${COMPILE_FLAGS} -O3) #-Wno-deprecated
endif()
```
If I remove the `-Werror` options here, then my build succeeds. Obviously we don't want to remove the flags here as we do want warnings as errors in Marabou itself. I'm about as far from a C++ expert as you can get, but this suggests to me that somehow these flags *are* getting propagated to the boost library? 

Random musings: perhaps the main Marabou build process is not linking against the built boost libraries, but is instead rebuilding them?

Anyway I'll press on for the moment locally using the temporary hack above.

MatthewDaggitt:
Hmm as an aside I've just tried downloading an entirely fresh copy of the repo and building the `master` branch, and I have exactly the same problem....

AleksandarZeljic:
> > Can you run that script in the tools folder and see if the issue persists? If it goes through without issues then: a) we indeed pass the flag somehow (which will be for me to figure out) and b) you will have boost installed and Marabou should not complain from that point onward.
> 
> So I ran `download-boost.sh` manually. That ran with no errors. I then went back to the `build` folder and ran `cmake --build .` as before, but still got the same error.
> 
> After a bit of fiddling I found that the place that is enabling the `werror` warning is the snippet below in the top-level `CMakeLists` file:
> 
> ```
> if (NOT MSVC)
>     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
>        set(COMPILE_FLAGS  -Wall -Wextra  -Werror -MMD -Qunused-arguments ) #-Wno-deprecated 
>      else()
>        set(COMPILE_FLAGS  -Wall -Wextra  -Werror -MMD ) #-Wno-deprecated 
>     endif()
>     set(RELEASE_FLAGS ${COMPILE_FLAGS} -O3) #-Wno-deprecated
> endif()
> ```
> 
> If I remove the `-Werror` options here, then my build succeeds. Obviously we don't want to remove the flags here as we do want warnings as errors in Marabou itself. I'm about as far from a C++ expert as you can get, but this suggests to me that somehow these flags _are_ getting propagated to the boost library?
> 
> Random musings: perhaps the main Marabou build process is not linking against the built boost libraries, but is instead rebuilding them?
> 
> Anyway I'll press on for the moment locally using the temporary hack above.

Thanks for the follow-up. In this case this is not a C++ issue, this is a CMake issue, which I am not extremely knowledgeable about myself. I will look into it further.

Regarding next steps, I have gone over your code and it is very tidy, modular and clean and I honestly don't have structural complaints. I will take another look to compare with the python implementation. 

In the mean time, we would want to add some test cases to Marabou that use the .onnx parser. We can do that by adding some .onnx networks to the resources folder and we could take networks from the VNNComp benchmarks that match some of the tests we have in .nnet format. For example, the acas benchmarks can be found here - https://github.com/stanleybak/vnncomp2021/tree/main/benchmarks/acasxu. Then we can simply add another say 10 or so instances that do not take too long to solve to the test suite. You are also welcome to include any of your own examples as tests, we will just need to classify them according to solving time. Ideally tests should be quick to execute. Please let me know if you need any help or if you have questions.

MatthewDaggitt:
Okay, I've added the ONNX ACASXu networks as suggested, and have tweaked the test setup so whenever you setup an ACASXu test it runs it both on the `.onnx` and the `.nnet` networks. Tests pass locally, but will continue to fiddle with the CI to get the tests to pass on Github. It would be great to add some more tests from the MarabouPy test suite, but they all use currently unimplemented ONNX operators. As I add support for more operators in future PRs, I'll add further tests from there as well.

@AleksandarZeljic let me know what other tweaks need to be made to the main `OnnxParser.cpp` file :+1:

MatthewDaggitt:
Okay, this latest error has defeated my CMake knowledge. Locally on my machine it successfully builds and finds the `onnx.proto3.pb.cc` file, however the CI is erroring with:
```
[ 11%] Linking CXX executable Marabou
/usr/bin/ld: libMarabouHelper.a(Marabou.cpp.o): in function `OnnxParser::~OnnxParser()':
Marabou.cpp:(.text._ZN10OnnxParserD2Ev[_ZN10OnnxParserD2Ev]+0x96): undefined reference to `onnx::GraphProto::~GraphProto()'
/usr/bin/ld: libMarabouHelper.a(OnnxParser.cpp.o): in function `OnnxParser::OnnxParser(String const&)':
OnnxParser.cpp:(.text+0x2a): undefined reference to `onnx::GraphProto::GraphProto(google::protobuf::Arena*, bool)'
/usr/bin/ld: OnnxParser.cpp:(.text+0x1ef): undefined reference to `onnx::ModelProto::ModelProto(google::protobuf::Arena*, bool)'
/usr/bin/ld: OnnxParser.cpp:(.text+0x216): undefined reference to `onnx::_GraphProto_default_instance_'
...
```

I suspect the culprit is the two lines I've added at lines 269-270 in the top-level `CMakeLists.txt` file. Maybe I should be linking them as a proper library in the same way that CVC4 is? Or are the parsers somehow built as a separate library and I need to add onnx as a dependency to that library?

AleksandarZeljic:
> I suspect the culprit is the two lines I've added at lines 269-270 in the top-level CMakeLists.txt file. Maybe I should be linking them as a proper library in the same way that CVC4 is? Or are the parsers somehow built as a separate library and I need to add onnx as a dependency to that library?

I could reproduce the issue locally. Based on the reported error it seems that the onnx_parser is not linked correctly. So all the parsers are built in the same way, and seemingly link against the MARABOU_LIB ( see the marabou_parser macro in src/input_parsers/CMakeLists.txt). It is possible that the onnx dependencies need to be linked there more explicitily  somehow. Let me know if this helps.

src/engine/MarabouError.h
```diff
@@ -58,6 +58,7 @@ class MarabouError : public Error
         UNSUPPORTED_PIECEWISE_LINEAR_CONSTRAINT = 102,
         UNSUPPORTED_TRANSCENDENTAL_CONSTRAINT = 103,
         UNSUPPORTED_NON_LINEAR_CONSTRAINT = 104,
+        ONNX_PARSE_ERROR = 105,
```
AleksandarZeljic, line 61: Should this be PARSER or PARSING?

src/input_parsers/NetworkParser.h
```diff
@@ -0,0 +1,70 @@
...
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Matthew Daggitt
+ ** This file is part of the Marabou project.
+ ** Copyright (c) 2017-2019 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved. See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** This file provides a general interface for parsing a neural network.
```
AleksandarZeljic, line 12: Can you elaborate on the class description here? What is the purpose of the NetworkParser and what format is it parsing?

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+ ** All rights reserved. See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** A parser for ONNX neural network files.
+ **
+ ** Some hard won knowledge about the design in this file:
+ **   - The data is annoyingly heterogenous. Inputs and outputs of the onnx graph
+ **   have different types from the nodes in the onnx graph, which have different
+ **   types from the initializers (i.e. the constants (biases, filter weights etc.).
+ **   This explains why all the methods takes the names of things as inputs
```
AleksandarZeljic, line 18: typo: takes->take

src/input_parsers/OnnxParser.h
```diff
@@ -0,0 +1,91 @@
...
+    void generateQuery( InputQuery &inputQuery );
+    void generatePartialQuery( InputQuery &inputQuery , Set<String> &inputNames, String &outputName );
+
+private:
+
+    onnx::GraphProto _network;
+    Set<String> _inputNames;
+    String _outputName;
+    Set<String> _processedNodes;
+    uint _numberOfFoundInputs;
```
AleksandarZeljic, line 40: The type here should be unsigned, since uint is not a standard type and its support is compiler dependent.

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+        newShape[inferredIndex] = tensorSize(oldShape) / tensorSize(newShape);
+    }
+    return newShape;
+}
+
+void checkEndianness()
+{
+    int num = 1;
+    bool systemIsLittleEndian = *(char *)&num == 1;
+    if (! systemIsLittleEndian)
```
AleksandarZeljic, line 252: The ! operator should not have a space sparating it from the argument. 

On a general note, we use a space inside brackets on each side (except for type casting). So that should be fixed globally throughout the PR.

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+}
+
+float getFloatAttribute(onnx::NodeProto& node, String name, float defaultValue)
+{
+    const onnx::AttributeProto* attr = findAttribute(node, name, onnx::AttributeProto_AttributeType_FLOAT);
+    if (attr == nullptr)
+    {
+        return defaultValue;
+    }
+    return (*attr).f();
```
AleksandarZeljic, line 282: The syntax shorthand we prefer for this expression is attr->f();

AleksandarZeljic: Also a number of instances below fit this pattern.

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+ */
+void OnnxParser::initialiseMaps()
+{
+    Set<std::string> names;
+    for (const onnx::NodeProto& node : _network.node())
+    {
+        const std::string name = node.name();
+        if (names.exists(name))
+        {
+            String errorMessage = "nodes in Onnx network must have a unique name";
```
AleksandarZeljic, line 337: Capitalize nodes in the error message. Also, below.

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+        {
+            message += " " + output.name();
+        }
+        throw MarabouError(MarabouError::ONNX_PARSE_ERROR, message.ascii());
+    }
+    std::cout << "\noutput: " + _network.output()[0].name();
+    return _network.output()[0].name();
+}
+
+void OnnxParser::initialiseShapeMap()
```
AleksandarZeljic, line 431: Can we use initialize here instead, for consistency with the rest of the code base?

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+    }
+}
+
+void OnnxParser::processGraph(Set<String> &inputNames, String &outputName, InputQuery& query)
+{
+    _inputNames = inputNames;
+    _outputName = outputName;
+
+    initialiseShapeMap();
+    processNode(outputName, true);
```
AleksandarZeljic, line 487: Does this assume that there is only one network output?

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+
+    List<onnx::NodeProto> nodes = getNodesWithOutput(nodeName);
+    assert(nodes.size() == 1);
+    onnx::NodeProto& node = nodes.front();
+
+    // First recursively process the input nodes.
+    // This ensures that shapes and values of a node's inputs have been computed first.
+    for (auto inputNode : getInputsToNode(node))
+    {
+        processNode(inputNode, makeEquations);
```
AleksandarZeljic, line 526: Does this need to be done in a recursive way? There will be many redundant calls to already processed inputs compared to an iterative inputs-to-outputs traversal of the graph which touches each node exactly once.

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+
+    _shapeMap[outputNodeName] = _shapeMap[inputNodeName];
+    if (_varMap.exists(inputNodeName))
+    {
+        _varMap[outputNodeName] = _varMap[inputNodeName];
+    }
+    else if (_constantMap.exists(inputNodeName))
+    {
+        _constantMap[outputNodeName] = _constantMap[inputNodeName];
+    }
```
AleksandarZeljic, line 702: Should an error be thrown otherwise?

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+ */
+void OnnxParser::reluEquations( onnx::NodeProto& node, bool makeEquations )
+{
+    String outputNodeName = node.output()[0];
+    String inputNodeName = node.input()[0];
+
+    _shapeMap[outputNodeName] = _shapeMap[inputNodeName];
+    if (!makeEquations)
+    {
+        return;
```
AleksandarZeljic, line 905: One liner bodies are allowed.

src/input_parsers/onnx_example/main.cpp
```diff
@@ -0,0 +1,30 @@
...
+ **/
+
+#include "Engine.h"
+#include "InputQuery.h"
+#include "MpsParser.h"
+#include <cstdio>
+
+int main( int argc, char *argv[] )
+{
+    return 0;
```
AleksandarZeljic, line 21: This could be nearly identical to the AcasParser, right?

AleksandarZeljic:
Hi @MatthewDaggitt, thank you for your patience, the remainder of the feedback is above. The only bigger thing is the recursive implementation of the graph traversal, which given the density of connections may cause many spurious calls and an iterative solution makes more sense. 

Let me know if anything needs clarifying or if I you need me to do more digging into the CMake situation.

src/engine/MarabouError.h
MatthewDaggitt: Changed to `PARSER`

src/input_parsers/OnnxParser.cpp
MatthewDaggitt: Fixed

src/input_parsers/OnnxParser.h
MatthewDaggitt: Fixed

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+}
+
+TensorShape shapeOfInput(onnx::ValueInfoProto &input)
+{
+    TensorShape result;
+    for (auto dim : input.type().tensor_type().shape().dim())
+    {
+        int size = dim.dim_value();
+        // This is copied from the Python code, but it feels like an
+        // error should be thrown if the size is non-positive.
```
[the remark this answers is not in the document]
MatthewDaggitt: Now throws error

src/input_parsers/OnnxParser.cpp
MatthewDaggitt: Fixed everywhere I could find

src/input_parsers/OnnxParser.cpp
MatthewDaggitt: Fixed in both places.

src/input_parsers/OnnxParser.cpp
MatthewDaggitt: Changed to `z` everywhere

src/input_parsers/OnnxParser.cpp
MatthewDaggitt: Yes, not a feature currently supported at the moment (was added to the Python version after I started this PR). Added a doc string to this function explaining this.

src/input_parsers/OnnxParser.cpp
MatthewDaggitt: To be honest I'm not sure, I think it depends on what constraints are on the topology of an ONNX graph? This is a straightforward port of what the existing python code does. 

I don't think the performance hit is worth worrying about though. Each revisited node does a constant time lookup in the `_processedNodes` set and is then is discarded when found.

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+/**
+ * @brief Compute the shape and values of a node assuming the input shapes
+ * and values have been computed already.
+ *
+ * @param node Name of node for which we want to compute the output shape
+ * @param makeEquations Create Marabou equations for this node if True
+ */
+void OnnxParser::makeMarabouEquations(onnx::NodeProto &node, bool makeEquations)
+{
+    printf("\nProcessing %s", node.name().c_str());
```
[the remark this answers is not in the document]
MatthewDaggitt: Done

src/input_parsers/OnnxParser.cpp
MatthewDaggitt: Changed everywhere

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1265 @@
...
+{
+    String nodeName = node.output()[0];
+
+    // Get inputs and determine which inputs are constants and which are variables
+    String input1Name = node.input()[0];
+    String input2Name = node.input()[1];
+    TensorShape input1Shape = _shapeMap[input1Name];
+    TensorShape input2Shape = _shapeMap[input2Name];
+
+    assert( input1Shape.last() == input2Shape.first() );
```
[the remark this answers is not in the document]
MatthewDaggitt: Changed everywhere

src/input_parsers/onnx_example/main.cpp
MatthewDaggitt: Well no because we can't evaluate the error as we can't run inference on the Onnx network (would need the `onnxruntime` library to do that).

Leaving that aside, I don't really understand the purpose of these example files? Why do they exist in addition to the test suite?

src/input_parsers/OnnxParser.cpp
MatthewDaggitt: This was honestly the most painful thing I've ever done in my 15 years of programming. Both aesthetically and practically. Couldn't find any C++ code formatter that supports such a style...

src/input_parsers/OnnxParser.cpp
MatthewDaggitt: Also how many nodes are actually in the average graph? 50 would be large right? I'm not imagining more than 200 for even the huge NLP networks?

MatthewDaggitt:
> The only bigger thing is the recursive implementation of the graph traversal, which given the density of connections may cause many spurious calls and an iterative solution makes more sense.

See my response above.

> Let me know if anything needs clarifying or if I you need me to do more digging into the CMake situation.

So can I ask you to try building the latest version locally again ?

The errors I get seem to depend on the placement of the ONNX library linking code (see [a8c5ba5](https://github.com/NeuralNetworkVerification/Marabou/pull/553/commits/a8c5ba5c32855e8d34b1421d500f83f0da3e3e50)).
 If I place it at line 264 in the `CMakeLists` file then I cannot build it locally, you cannot build it locally and the CI cannot build it. They all error at 11% when trying to link the Marabou binary. 

However if I place it at line 61 then I _can_ build it locally, but the CI cannot build it failing at 3% saying they can't even find the header files. Can you build it locally?

src/input_parsers/OnnxParser.cpp
AleksandarZeljic: From a C++ perspective, every function call is a new stack frame that can be avoided in the reverse traversal, especially since its made only for a map lookout. For example the VNN-Comp uses 5minutes timeout per benchmark, any time not spent parsing is time that can be spent solving the benchmarks. 

So a fully connected network of say 50 nodes per layer will make 50x50=2500 calls recursive calls per layer only to perform a map look up, and 2450 of those will be spurious (since the first node in the layer will already recursively pre-process all the inputs). For 200 nodes per layer that would be 40000 - 200 spurious calls per layer. In any case, this is probably good enough for now. The recursive to iterative solution is common pattern and simple enough to be done after the fact.

src/input_parsers/OnnxParser.cpp
AleksandarZeljic: Sounds good.

src/input_parsers/OnnxParser.cpp
AleksandarZeljic: Oh, wow, thanks for doing that! Clang-format has SpacesInParentheses option that does that. I am sorry, I didn't think of providing a pointer ahead of time.

AleksandarZeljic:
Thanks for addressing the feedback and apologies for sending you on a manual goose chase. 

Like you predicted, it fails at 3% locally on my machine. I remember similar issues when adding the CVC4 dependencies. It was crucial to find the right spot to include them in the build process, but in the end there was no clear reason why it worked when it finally did.

MatthewDaggitt:
Okay, I think I've managed to fix it! The tests are passing :tada: 

I think all the issues are resolved. I've rebased on `master` as you requested previously. Would you like me to squash all the commits down into one? I don't think it should go in as is.

src/input_parsers/OnnxParser.cpp
AleksandarZeljic: Oh, got it. I spent too much time with explicit representations for my own good it seems. Yeah, no issues then! Thanks for the clarification.

AleksandarZeljic (approved):

AleksandarZeljic:
Great, thanks for all your hard work on this! Github will squash them for us when merging.

AleksandarZeljic:
What platform did you develop this on? I want to have someone with MacOS build it just in case, since there's a lot of new things added to the code base. I will delay the merge until then.

MatthewDaggitt:
> What platform did you develop this on? I want to have someone with MacOS build it just in case, since there's a lot of new things added to the code base. I will delay the merge until then.

Ubuntu 20.04 I'm afraid. Okay, looking forward to seeing the back of this one :smile:

src/input_parsers/OnnxParser.cpp
```diff
@@ -0,0 +1,1281 @@
...
+}
+
+Set<String> OnnxParser::readInputNames()
+{
+    ASSERT( _network.input().size() >= 1 );
+
+    Set<String> initializerNames;
+    for ( auto &initNode : _network.initializer() )
+    {
+        std::cout << "\ninitialiser: " + initNode.name();
```
anwu1219, line 402: Either remove the print statements or use the "ONNX_LOG" instead so that these messages can be seen only when debugging?

anwu1219:
Hi @MatthewDaggitt , thanks a lot for this contribution! I just tried to build the branch on Mac, and it seems that when the protobuf build is being configured, the wrong compiler can be selected. This can be fixed by adding these lines to download_protobuf.sh (say at line 11): https://github.com/NeuralNetworkVerification/Marabou/blob/14ea65476e25dfa6430839e4d6a2039c08881628/tools/download_openBLAS.sh#L14-L16

After this patch, I can build and execute the tool without problems. When trying out the parser, I notice that there are outputs such as:

```
Onnx network!

initialiser: input_AvgImg
initialiser: Operation_1_MatMul_W
initialiser: Operation_1_Add_B
initialiser: Operation_2_MatMul_W
initialiser: Operation_2_Add_B
...
```

Perhaps we want to either remove them or use the ONNX_LOG method? I left a comment regarding this.

anwu1219:
Another comment is that we unfortunately have two entry points. One for single threaded mode, another for parallel mode (either split-and-conquer or a "portfolio" mode). Currently only the former is updated. Could you please also copy this new logic: https://github.com/MatthewDaggitt/Marabou/blob/0dfc3887f70ecdc6476f052d04aadaa631f17b9c/src/engine/Marabou.cpp#L103-L108

to this file?
https://github.com/MatthewDaggitt/Marabou/blob/0dfc3887f70ecdc6476f052d04aadaa631f17b9c/src/engine/DnCMarabou.cpp#L66

Currently when I try to execute the parallel mode: 
`./Marabou ../resources/onnx/acasxu/ACASXU_experimental_v2a_2_9.onnx --verbosity=0 --num-workers 8 ../resources/properties/acas_property_3.txt `

I get a segmentation fault. 

While if I do: 
`./Marabou ../resources/nnet/acasxu/ACASXU_experimental_v2a_2_9.nnet --verbosity=0 --num-workers 8 ../resources/properties/acas_property_3.txt `

The solver terminates with "unsat". 

I'm sorry for the late additional comments, but I think this is all for now.

MatthewDaggitt:
Thanks for the review @anwu1219 !

> When trying out the parser, I notice that there are outputs such as:

Thanks, nice spot. Didn't mean to leave those in. Have fixed.

> Another comment is that we unfortunately have two entry points. 

Have done so. Can I ask why there are two separate entry points? If there's a good reason, it would probably be good to add some documentation to both of them pointing at the other one?

> Currently when I try to execute the parallel mode:

I think the two commands you've given are identical? Regardless I think I've hopefully fixed it now.

anwu1219:
Hi @MatthewDaggitt , thanks for the patches! The PR looks good to me. 

> Why there are two separate entry points?

The main.cpp will decide whether to invoke the run() method in Marabou.h or DnCMarabou.h depending on the number of threads given. DnCMarabou will maintain a main thread that monitors the worker threads, while Marabou only has one threads. Ideally indeed it's probably better to branch after the network parsing instead of before.

> I think the two commands you've given are identical? 

One uses the network in .onnx format and one in .nnet format.

MatthewDaggitt:
Thanks for the explanation! I think I've fiddled with enough things in this PR as it is, so if it's okay with you I might leave refactoring the common code to `main.cpp` in a second PR. 

The command given by @anwu1219 above now works instead of segfaulting, so I think this is ready to merge @AleksandarZeljic?

AleksandarZeljic:
LGTM, thanks again @MatthewDaggitt! This is a really valuable contribution!

PR #557, opened 2022-04-07
Update RowBoundTightener to use BoundManager

AleksandarZeljic:
This PR replaces local bounds in the RowBoundTightener with BoundManager object, which loans out its local bounds that are context-dependent.


src/engine/Engine.cpp
```diff
@@ -745,14 +745,18 @@ bool Engine::performSimplexStep()
...
     }
 
     // Perform the actual pivot
     _activeEntryStrategy->prePivotHook( _tableau, fakePivot );
     _tableau->performPivot();
     _activeEntryStrategy->postPivotHook( _tableau, fakePivot );
 
+    if ( !fakePivot )
+    {
+        _rowBoundTightener->examinePivotRow();
```
anwu1219, line 757: Is it necessary now to move the examinePivotRow() call down here?

src/engine/RowBoundTightener.cpp
```diff
@@ -518,7 +433,25 @@ unsigned RowBoundTightener::tightenOnSingleConstraintRow( unsigned row )
       We first compute the lower and upper bounds for the expression
 
           sum ci xi - b
-    */
+          if ( FloatUtils::lt( _lowerBounds[xi], lowerBound ) )
```
anwu1219, line 436: Remove these?

anwu1219 (approved):

PR #560, opened 2022-04-29
Remove ConstraintBoundTightener

AleksandarZeljic:
This PR removes ConstraintBoundTightener class, since its features are now built-into the BoundManager. BoundManager uses _tightened* flags to keep track of recent tightenings, which are reset when tightenings are obtained using BoundManager::getBoundTightenings() method. This is functionally equivalent to ConstraintBoundTightener.

To make the transition this PR:

- Replaces ConstraintBoundTightener with BoundManager class
- Updates notifyLower/UpperBound methods in classes to reflect that BoundManager will already have up-to-date values.
- Updates tests to use MockBoundManager instead of ConstraintBoundTightener.
- Removes an overly strong SoI assertion


src/engine/tests/Test_ReluConstraint.h
```diff
@@ -1,5 +1,5 @@
 /*********************                                                        */
-/*! \file Test_ReluConstraint.h
+/*! \file _Test_ReluConstraint.h
```
anwu1219, line 2: Revert this change?

src/engine/tests/Test_SignConstraint.h
```diff
@@ -356,17 +357,17 @@ class SignConstraintTestSuite : public CxxTest::TestSuite
...
 
         splits = sign.getCaseSplits();
         TS_ASSERT_EQUALS( splits.size(), 2U );
 
         sign.notifyLowerBound( 4, -0.5 );
-        TS_ASSERT_THROWS_EQUALS( splits = sign.getCaseSplits(),
-                                 const MarabouError &e,
-                                 e.getCode(),
-                                 MarabouError::REQUESTED_CASE_SPLITS_FROM_FIXED_CONSTRAINT );
+        /* TS_ASSERT_THROWS_EQUALS( splits = sign.getCaseSplits(), */
```
anwu1219, line 367: Are we not throwing this error anymore?

anwu1219 (approved):

Issue #571, opened 2022-06-14
Runtime Error when using Gurobi and Bound tightening


marcel-v:
Hi,

I observe an runtime error when running Marabou with this [query](https://github.com/NeuralNetworkVerification/Marabou/files/8900700/query.txt) when using _--lp-solver gurobi_ and any bound tightening strategy, i.e., **not** using _--tightening-strategy none_.

I can trace the error to a bad access in `Tableau::updateVariableToComplyWithUpperBoundUpdate( unsigned int, double )`, 
specifically, `_variableToIndex[variable]` (Tableau.cpp:1770),
see the [stack](https://github.com/NeuralNetworkVerification/Marabou/files/8900770/stack.txt).
At access time, `_variableToIndex` still points to `NULL`.

Commit is **492c1b8c703c8a383f421468a104c34710e6d26d**.

Thanks in advance for any feedback!

Best,
Marcel

anwu1219:
Hi @marcel-v , thanks for your interest in the tool and tracking down the error. This does seem to be a issue stemming from this latest commit. 
We will fix the bug asap. 
For now, could you please revert back to the previous commit 2641ce74ccf0ebb53345c4a0f6b4c773df40e8a4? This one should work.

anwu1219:
Hi @marcel-v , please try the latest master. should be fixed.

marcel-v:
> Hi @marcel-v , please try the latest master. should be fixed.

Great, works!

marcel-v:
There appears to be another invalid access to the tableau internals when using Gurobi with [this query](https://github.com/NeuralNetworkVerification/Marabou/files/8924583/query.txt), resulting in
[this stack](https://github.com/NeuralNetworkVerification/Marabou/files/8924585/stack.txt).

I am at latest commit  eb93d44e66aec9e374691afc31cd2b3916d76a20.

anwu1219:
Hi @marcel-v , I apologize for the inconvenience. A new Pr should fix these issues, and all the regression tests now pass with Gurobi on.

PR #572, opened 2022-06-14
Prevent access to Tableau internals when not in native mode

AleksandarZeljic:
In Tableau::updateVariableToComplyWith*Bounds methods


anwu1219:
Tried on the query from issue https://github.com/NeuralNetworkVerification/Marabou/issues/571 and seems to fix the issue.

anwu1219 (approved):

PR #574, opened 2022-06-17
Add basic Softmax class

anwu1219:
- Add basic SoftmaxConstraint class (currently the only way to solve the a query with Softmax is through preprocessing).
- rename TranscendentalConstraint interface to NonlinearConstraint
- Add supportVariableElimination() method in PLConstraint and NonlinearConstraint (Disjunction and Sigmoid currently does not support variable elimination)
- Change inputOutputVariable in preprocessor


maraboupy/MarabouCore.cpp
```diff
@@ -97,8 +98,8 @@ void addReluConstraint(InputQuery& ipq, unsigned var1, unsigned var2){
 }
 
 void addSigmoidConstraint(InputQuery& ipq, unsigned var1, unsigned var2){
-    TranscendentalConstraint* s = new SigmoidConstraint(var1, var2);
-    ipq.addTranscendentalConstraint(s);
+    NonlinearConstraint* s = new SigmoidConstraint(var1, var2);
```
guykatzz, line 101: * s --> *s

maraboupy/MarabouCore.cpp
```diff
@@ -114,6 +115,18 @@ void addMaxConstraint(InputQuery& ipq, std::set<unsigned> elements, unsigned v){
     ipq.addPiecewiseLinearConstraint(m);
 }
 
+void addSoftmaxConstraint(InputQuery& ipq, std::list<unsigned> inputs,
```
guykatzz, line 118: Use List instead of std::list?

guykatzz: Also, add spaces to the parameter lists
void addSoftmaxConstraint( InputQuery& ipq, std::list<unsigned> inputs,
                                             ^

src/engine/Engine.h
```diff
@@ -279,7 +279,7 @@ class Engine : public IEngine, public SignalHandler::Signalable
     /*
       The existing transcendental constraints.
     */
-    List<TranscendentalConstraint *> _tsConstraints;
+    List<NonlinearConstraint *> _tsConstraints;
```
guykatzz, line 282: _tsConstraints should probably be renamed, too

src/engine/SoftmaxConstraint.cpp
```diff
@@ -0,0 +1,291 @@
...
+        tightenings.append( Tightening( _outputs[i], 1, Tightening::UB ) );
+
+        // First get tighter lower bound
+        bool validBound = true;
+        Vector<double> values;
+        for ( unsigned j = 0; j < _inputs.size(); ++j )
+        {
+            if ( i == j )
+            {
+                if ( existsLowerBound( _inputs[j] ) &&
```
guykatzz, line 188: I didn't understand what was going on here. Are we just taking the existing bounds and propagating them as is?

maraboupy/MarabouCore.cpp
anwu1219: Since we are going to call this function in python using pybind, I think we have to use std::list<unsigned>

anwu1219: The format in MarabouCore.cpp is completely different from the rest of the code base btw. Probably need a clang pass to fix those.

src/engine/Engine.h
anwu1219: done

PR #580, opened 2022-07-01
Transition to context dependent search state

AleksandarZeljic:
This PR transitions to using Context dependent architecture to store the search state.
This is achieved by using the CD features of PiecewiseLinearConstraints to store their current state in the search.
The SmtCore now uses the PiecewiseLinearConstraint::nextFeasibleCase/markUnfeasible to record the advancement of the search with the PiecewiseLinearConstraints.

This PR:
* Replaces old SmtCore with the contextDependent implementation which should be mostly equivalent.


src/engine/DnCWorker.cpp
```diff
@@ -101,7 +101,7 @@ void DnCWorker::popOneSubQueryAndSolve( bool restoreTreeStates )
 
         bool fullSolveNeeded = true; // denotes whether we need to solve the subquery
         if ( restoreTreeStates && smtState )
-            fullSolveNeeded = _engine->restoreSmtState( *smtState );
+            fullSolveNeeded = true; //_engine->restoreSmtState( *smtState );
```
anwu1219, line 104: Do we want to fully remove this functionality for now? Maybe we want to also purge this option here: https://github.com/NeuralNetworkVerification/Marabou/blob/bdcd044e6d64902d209ac5391285bb31a43662c5/src/configuration/OptionParser.cpp#L148?

src/engine/MaxConstraint.cpp
```diff
@@ -669,6 +669,7 @@ String MaxConstraint::serializeToString() const
         // Will be ignored in any case
         output += Stringf( ",%u", 0 );
 
+    //serializeInfeasibleCases( output );
```
anwu1219, line 672: Remove this line?

src/engine/PiecewiseLinearConstraint.cpp
```diff
@@ -168,13 +168,16 @@ void PiecewiseLinearConstraint::markInfeasible(
     _cdInfeasibleCases->push_back( infeasibleCase );
 }
 
-PhaseStatus PiecewiseLinearConstraint::nextFeasibleCase()
+PhaseStatus PiecewiseLinearConstraint::nextFeasibleCase() const
 {
-    ASSERT( getPhaseStatus() == PHASE_NOT_FIXED );
+    //ASSERT( getPhaseStatus() == PHASE_NOT_FIXED );
```
anwu1219, line 173: Remove this line?

src/engine/PiecewiseLinearConstraint.h
```diff
@@ -514,6 +523,7 @@ class PiecewiseLinearConstraint : public ITableau::VariableWatcher
      */
     bool isCaseInfeasible( PhaseStatus phase ) const;
 
+
```
anwu1219, line 526: Remove empty line

src/engine/oldSmtCore
```diff
@@ -0,0 +1,257 @@
+/*********************                                                        */
```
anwu1219, line 1: Remove this and oldSmtCorecpp?

src/engine/PiecewiseLinearConstraint.cpp
```diff
@@ -201,6 +204,35 @@ void PiecewiseLinearConstraint::setStatistics( Statistics *statistics )
     _statistics = statistics;
 }
 
+bool PiecewiseLinearConstraint::phaseFixed() const
```
anwu1219, line 207: When _cdPhaseStatus is not null, does this method always have the same output as isImplication()? If so do we want to merge the two methods?

src/engine/tests/Test_DisjunctionConstraint.h
```diff
@@ -205,23 +208,24 @@ class DisjunctionConstraintTestSuite : public CxxTest::TestSuite
             dc.notifyLowerBound( 2, -10 );
             dc.notifyUpperBound( 2, 10 );
 
-            TS_ASSERT( !dc.phaseFixed() );
+            TS_ASSERT( !dc.isImplication() );
```
anwu1219, line 211: Do we want to also check phaseFixed() iff isImplication()?

src/engine/DnCWorker.cpp
AleksandarZeljic: Good point, I will remove it altogether as it is obsolete now.

src/engine/tests/Test_DisjunctionConstraint.h
AleksandarZeljic: No the semantics actually differ. Disjunction does not ever explicitly fix the phase, it actually used a test that is exactly the implementation of isImplication. Therefore I removed the special implementation of DisjunctionConstraint::phaseFixed() and replaced the calls with isImplication().

src/engine/PiecewiseLinearConstraint.cpp
AleksandarZeljic: Longterm we want to merge the two methods. This PR already does a lot and I didn't want to complicate things further, since I am not convinced that this would be a trivial change (e.g. Max/Disjunction semantics). PhaseFixed currently is derived from bound propagation while isImplication is derived from search/backtracking.

PR #585, opened 2022-08-03
Support some ONNX operators

tagomaru:

This PR is going to support some ONNX operators.


maraboupy/MarabouNetworkONNX.py
```diff
@@ -23,6 +23,7 @@
 from maraboupy import MarabouNetwork
 from onnx import TensorProto
 import itertools
+import torch
```
anwu1219, line 26: Why do we need torch here?

maraboupy/MarabouNetworkONNX.py
```diff
@@ -761,13 +768,165 @@ def matMulEquations(self, node, makeEquations):
...
+        if not makeEquations:
+            return
+
+        # Get variables
+        inputVars = np.concatenate(inputVars, axis).reshape(-1)
+        outputVars = self.makeNewVariables(nodeName).reshape(-1)
+
+        assert len(inputVars) == len(outputVars)
+
+        for i in range(len(inputVars)):
```
anwu1219, line 801: For these operations, I don't think we need to actually introduce new equations or variables, which makes the verification query harder. Please the reshape, flatten, or transpose methods in this class.

maraboupy/MarabouNetworkONNX.py
tagomaru: Yes, it is used.

maraboupy/MarabouNetworkONNX.py
tagomaru: fixed.

maraboupy/MarabouNetworkONNX.py
```diff
@@ -761,13 +768,149 @@ def matMulEquations(self, node, makeEquations):
...
+        Args:
+            node (node): ONNX node representing the Resize operation
+            makeEquations (bool): True if we need to create new variables and write Marabou equations
+
+        :meta private:
+        """
+        nodeName = node.output[0]
+        inputName = node.input[0]
+
+        # Check numper of dimension of input
```
anwu1219, line 853: number

maraboupy/MarabouNetworkONNX.py
```diff
@@ -761,13 +768,149 @@ def matMulEquations(self, node, makeEquations):
...
+        inputVars = self.varMap[inputName]
+        inputShape = inputVars.shape
+        if inputVars.ndim != 4:
+            raise NotImplementedError("Marabou only supports resize operator for very specific upsample case used in YOLO now.")
```
anwu1219, line 857: Could you add a comment here describing the currently unsupported functionality of resize?

anwu1219 (approved):

maraboupy/MarabouNetworkONNX.py
tagomaru: fixed

maraboupy/MarabouNetworkONNX.py
tagomaru: added

PR #586, opened 2022-09-01
Add export/import Assignment features for debugging

AleksandarZeljic:
This PR exposes functionality to debug a solution to the command line. See --help for detailed command line arguments.
This feature is intended for developers, but can be used to report bugs by the users as well. 
The basic use case scenario: a satisfying assignment is cut off (unsoundly) during the search on the development branch. 
The user can then export the satisfying assignment using the sound version of the tool and import it into the unsound version to obtain information at which point in the search is the satisfying assignment cut off.


anwu1219 (approved):
LGTM!

PR #587, opened 2022-09-15
Add Statistic tracking for Context push/pop operations

AleksandarZeljic:
This PR adds basic statistic tracking of number of context push/pop operations, as well as time breakdown between the push/pop and related hook calls.

### DIFF
--- a/src/common/Statistics.cpp
+++ b/src/common/Statistics.cpp
@@ -29,6 +29,8 @@ Statistics::Statistics()
     _unsignedAttributes[MAX_DECISION_LEVEL] = 0;
     _unsignedAttributes[NUM_SPLITS] = 0;
     _unsignedAttributes[NUM_POPS] = 0;
+    _unsignedAttributes[NUM_CONTEXT_PUSHES] = 0;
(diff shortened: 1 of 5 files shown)


src/common/Statistics.cpp
```diff
@@ -391,6 +397,53 @@ void Statistics::print()
             , totalTimeGettingSoIPhasePatternMicro
             , printPercents( totalTimeGettingSoIPhasePatternMicro,
                              timeMainLoopMicro ) );
+
+    printf( "\t--- Context dependent statistics ---\n" );
```
anwu1219, line 401: We printed NUM_POPS and NUM_SPLITS above. How are NUM_CONTEXT_PUSHES/POPS different from those? (Do we want to print these statistics together?)

AleksandarZeljic: The semantics actually differ, since push and pop above are the old SmtCore implementation. I printed them here, since these are actual context operations and they can be used to gauge the average cost of a push/pop operation. In the new SmtCore the two semantics will actually match and I would still want the statistics shown in both places. Maybe even introduce average calculation or something.

anwu1219 (approved):

PR #588, opened 2022-09-15
export preprocess query to maraboupy

elazarcoh:
This add a way to get the current query with the current bound, allowing to return a processed query to maraboupy.


src/engine/Engine.cpp
```diff
@@ -3089,3 +3089,13 @@ bool Engine::consistentBounds() const
 {
     return _boundManager.consistentBounds();
 }
+
```
guykatzz, line 3092: remove one of the empty lines

guykatzz (approved):
See one small comment

PR #591, opened 2022-09-29
Refactor shared InputQuery preparation into `main.cpp`

MatthewDaggitt:
Follows up on the suggestion by @anwu1219 to refactor the input query parsing into `main.cpp` before it branches on whether to use `Marabou` or `DNCMarabou`.

### DIFF
--- a/src/engine/DnCMarabou.cpp
+++ b/src/engine/DnCMarabou.cpp
@@ -25,7 +25,7 @@
 #include "QueryLoader.h"
 #include "AcasParser.h"
 
-DnCMarabou::DnCMarabou( InputQuery& inputQuery )
+DnCMarabou::DnCMarabou( InputQuery &inputQuery )
     : _inputQuery( inputQuery )
     , _dncManager( nullptr )
 {
--- a/src/engine/DnCMarabou.cpp
+++ b/src/engine/DnCMarabou.cpp
@@ -25,83 +25,16 @@
 #include "QueryLoader.h"
 #include "AcasParser.h"
 
-DnCMarabou::DnCMarabou()
-    : _dncManager( nullptr )
-    , _inputQuery( InputQuery() )
+DnCMarabou::DnCMarabou( InputQuery& inputQuery )
+    : _inputQuery( inputQuery )
+    , _dncManager( nullptr )
 {
+
 }
 
 void DnCMarabou::run()
 {
-    String inputQueryFilePath = Options::get()->getString( Options::INPUT_QUERY_FILE_PATH );
-    if ( inputQueryFilePath.length() > 0 )
-    {
-        /*
-          Step 1: extract the query
-        */
-        if ( !File::exists( inputQueryFilePath ) )
-        {
-            printf( "Error: the specified inputQuery file (%s) doesn't exist!\n", inputQueryFilePath.ascii() );
-            throw MarabouError( MarabouError::FILE_DOESNT_EXIST, inputQueryFilePath.ascii() );
-        }
-
-        printf( "InputQuery: %s\n", inputQueryFilePath.ascii() );
-        _inputQuery = QueryLoader::loadQuery( inputQueryFilePath );
-        _inputQuery.constructNetworkLevelReasoner();
-    }
-    else
-    {
-        /*
-          Step 1: extract the network
-        */
-        String networkFilePath = Options::get()->getString( Options::INPUT_FILE_PATH );
-        if ( !File::exists( networkFilePath ) )
-        {
-            printf( "Error: the specified network file (%s) doesn't exist!\n",
-                    networkFilePath.ascii() );
-            throw MarabouError( MarabouError::FILE_DOESNT_EXIST,
-                                networkFilePath.ascii() );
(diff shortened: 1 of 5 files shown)


src/engine/Marabou.cpp
```diff
@@ -30,33 +30,23 @@
...
-    {
-        delete _acasParser;
-        _acasParser = NULL;
-    }
-
-    if ( _onnxParser )
-    {
-        delete _onnxParser;
-        _onnxParser = NULL;
-    }
```
MatthewDaggitt, line 52: I'm not quite sure if I need to free the `_inputQuery` object here? Would appreciate the input of a slightly better C++ programmer!  (and if so what is the best syntax to use?)

src/engine/Marabou.cpp
```diff
@@ -30,33 +30,23 @@
...
 }
 
 void Marabou::run()
 {
+    if ( Options::get()->getBool( Options::DEBUG_ASSIGNMENT ) )
+        importDebuggingSolution();
+
     struct timespec start = TimeUtils::sampleMicro();
 
-    prepareInputQuery();
```
MatthewDaggitt, line 59: So the one change that I can see is that parsing the input query is no longer timed. If that is important I can add a separate timing function to time it?

MatthewDaggitt:
@anwu1219  any chance for a review of this?

anwu1219:
> @anwu1219 any chance for a review of this?

Hi Matthew, thanks a lot for the PR. I'll take a look within a couple of day.

src/engine/DnCMarabou.cpp
```diff
@@ -25,83 +25,16 @@
 #include "QueryLoader.h"
 #include "AcasParser.h"
 
-DnCMarabou::DnCMarabou()
-    : _dncManager( nullptr )
-    , _inputQuery( InputQuery() )
+DnCMarabou::DnCMarabou( InputQuery& inputQuery )
```
anwu1219, line 28: Style change:

( InputQuery& inputQuery ) 
=> 
(InputQuery &inputQuery)

src/engine/DnCMarabou.h
```diff
@@ -23,16 +23,16 @@
...
-    DnCMarabou();
+    DnCMarabou( InputQuery& inputQuery );
 
     /*
       Entry point of this class
     */
     void run();
 
 private:
+    InputQuery& _inputQuery;
```
anwu1219, line 34: Style change:

InputQuery& _inputQuery; 
=> 
InputQuery &_inputQuery;

src/engine/main.cpp
```diff
@@ -76,18 +159,28 @@ int main( int argc, char **argv )
             return 0;
         };
 
+        InputQuery inputQuery = prepareInputQuery();
```
anwu1219, line 162: To avoid unnecessary calls of the copy constructor, maybe declare the prepareInputQuery method like this:

`void prepareInputQury(InputQuery &inputQuery);
`
The method will update the given reference of the inputQuery object, which is created in the main function, like this:

```
InputQuery inputQuery; 
prepareInputQuery(inputQuery);

```

src/engine/Marabou.cpp
anwu1219: I think this is fine for now. Maybe add a TODO in the comment so we can revisit it later?

anwu1219 (changes requested):
Thanks for this. The changes look good overall. Just one suggested change that will avoid one copy of the InputQuery object, could make a difference when the network is large.

src/engine/main.cpp
MatthewDaggitt: Hmm so I've tried to do this, but it's pretty tricky. In particularly we eventually end up having to push that change all the way through to the method `QueryLoader.loadQuery`. From there, that method then gets called via the Python/C interface which I have no idea how to update properly.

I've added a comment to the effect that this change _should_ be done. Is there anything else I can do to improve this without meddling with the Python/C interface?

MatthewDaggitt:
Otherwise I've addressed all your comments, and, as requested in the previous PR, have rebased and then forced pushed.

MatthewDaggitt:
Can't reproduce the bug locally, so pushed an empty commit to retrigger the CI and see if it was just a CI glitch.

MatthewDaggitt:
Okay so this test failure is definitely not this PRs fault Same test is failing in https://github.com/NeuralNetworkVerification/Marabou/pull/627 in a totally unrelated PR.

@anwu1219 any ideas?

src/engine/main.cpp
anwu1219: Hum.. Ideally this change can be pushed through because we don't want to make unnecessary copies. 
This is not an issue for small networks, but for larger network where one only wants to do bound propagation on this without search, copying the input query can become a runtime and memory bottleneck.

anwu1219:
> Okay so this test failure is definitely not this PRs fault Same test is failing in #627 in a totally unrelated PR.
> 
> @anwu1219 any ideas?

This is very wierd. I should take a look at this regression test.

PR #619, opened 2023-02-01
Fix bug in SmtLibWriter

OmriIsacHUJI:


src/common/MString.cpp
```diff
@@ -175,6 +175,23 @@ String String::trim() const
     return substring( firstNonSpace, lastNonSpace - firstNonSpace + 1 );
 }
 
+String String::trimZerosFromRight() const
```
guykatzz, line 178: Wouldn't it make more sense to traverse the string from the right?

src/proofs/SmtLibWriter.cpp
```diff
@@ -91,5 +92,8 @@ void SmtLibWriter::writeInstanceToFile( IFile &file, const List<String> &instanc
 
 String SmtLibWriter::signedValue( double val )
 {
-    return val > 0 ? std::to_string( val ) : "( - " + std::to_string( abs( val ) ) + " )";
+    unsigned precision = ( unsigned ) std::log10 ( 1 / GlobalConfiguration::DEFAULT_EPSILON_FOR_COMPARISONS );
+    std::stringstream s;
+    s << std::fixed << std::setprecision( precision ) << abs( val );
+    return val >= 0 ? String ( s.str() ).trimZerosFromRight() : String( "( - " + s.str() ).trimZerosFromRight() + " )";
```
guykatzz, line 98: No space between a function name and the ensuing (
String ( s.str() --> String( s.str()
Likewise with std::log10 ( 1 ...

guykatzz (commented):
Also, I believe this PR fails the integration tests

src/common/MString.cpp
omriisack: It will. Fixed.

guykatzz (approved):

PR #627, opened 2023-03-28
Fix bug when parsing unknown dimension sizes in `OnnxParser.cpp`

MatthewDaggitt:
Hopefully self-explanatory given code documentation.


MatthewDaggitt:
Okay so this test failure is definitely not this PRs fault as that test doesn't even use the `OnnxParser`. Same test is failing in https://github.com/NeuralNetworkVerification/Marabou/pull/591. 

@anwu1219 any ideas?

Issue #625, opened 2023-03-24
Property parser doesn't parse non-simplified bound


MatthewDaggitt:
Feeding in the following property:
```
60261.0x0 >= 36156.6
```
results in an empty error message:
```
  Network: ACASXU_run2a_1_8.onnx
  Property: /tmp/specification-990a1536460c78d9/property2-query1.txt
  Caught a InputParserError error. Code: 1, Errno: 0, Message: .
```

This appears to be falling foul of this special case in the property parser that doesn't handle coefficients properly:

https://github.com/NeuralNetworkVerification/Marabou/blob/663835df7b56be42af94b3d2723bf1b69494c0d5/src/input_parsers/PropertyParser.cpp#L80

PR #628, opened 2023-03-31
Refactored PropertyParser and improved error messages

MatthewDaggitt:
This is a no-op refactoring of `PropertyParser` that unifies the variable parsing code between the case when we're parsing a bound and the case when we're parsing a full equation. This is in preparation for fixing #625, but that fix is not in this PR.

The only piece of functionality change is that the parser now skips empty lines.


MatthewDaggitt:
Closing for the moment, as I need to rethink this.

PR #631, opened 2023-04-12
Fix bug in ONNX parser with network with duplicate names

MatthewDaggitt:
Node names are not necessarily unique, and actually don't need to be for the parser code to work. Only input node names need to be unique


MatthewDaggitt:
Same non-deterministic test error as in https://github.com/NeuralNetworkVerification/Marabou/pull/627 and https://github.com/NeuralNetworkVerification/Marabou/pull/591

anwu1219 (approved):
LGTM

Issue #638, opened 2023-05-16
Segfault when solving for network with single sigmoid layer 


MatthewDaggitt:
Follow up from #630. When running the following (currently disabled) test:

https://github.com/NeuralNetworkVerification/Marabou/blob/fc0400c2f561393a98b95a2d2d1016a6bbbabb9a/src/input_parsers/tests/Test_OnnxParser.h#L267

the following error occurs:
```
Running 1 testAddressSanitizer:DEADLYSIGNAL
=================================================================
==1511045==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x557671a36fd6 bp 0x7fff1c539f20 sp 0x7fff1c539ec0 T0)
==1511045==The signal is caused by a READ memory access.
==1511045==Hint: address points to the zero page.
    #0 0x557671a36fd6 in Tableau::storeState(TableauState&, TableauStateStorageLevel) const /home/matthew/Code/AISEC/Marabou/src/engine/Tableau.cpp:1632
    #1 0x5576718ff041 in Engine::storeState(EngineState&, TableauStateStorageLevel) const /home/matthew/Code/AISEC/Marabou/src/engine/Engine.cpp:1699
    #2 0x55767199a657 in PrecisionRestorer::storeInitialEngineState(IEngine const&) /home/matthew/Code/AISEC/Marabou/src/engine/PrecisionRestorer.cpp:27
    #3 0x5576718d8b45 in Engine::storeInitialEngineState() /home/matthew/Code/AISEC/Marabou/src/engine/Engine.cpp:2194
    #4 0x5576718fa592 in Engine::solve(unsigned int) /home/matthew/Code/AISEC/Marabou/src/engine/Engine.cpp:188
    #5 0x5576718cb017 in OnnxParserTestSuite::run_test(String, Vector<double>, Vector<double>) (/home/matthew/Code/AISEC/Marabou/build/tests/parser/Test_OnnxParser+0x54017)
    #6 0x5576718cce64 in OnnxParserTestSuite::test_sigmoid() (/home/matthew/Code/AISEC/Marabou/build/tests/parser/Test_OnnxParser+0x55e64)
    #7 0x5576718cd403 in TestDescription_OnnxParserTestSuite_test_sigmoid::runTest() (/home/matthew/Code/AISEC/Marabou/build/tests/parser/Test_OnnxParser+0x56403)
[...]

PR #656, opened 2023-07-21
Add support for tanh to C++ OnnxParser

MatthewDaggitt:
Can't run the unit test, as `tanh` is really a `sigmoid` node which is susceptible to #638, but can confirm from my own larger tests that it works correctly.
 
As we now have versioning, I've also taken the liberty of adding a CHANGELOG entry.


wenkokke:
From the diff, it looks like we should configure git to treat onnx files as binary data.

wu-haoze (commented):
LGTM

wu-haoze (approved):

Issue #666, opened 2023-08-22
Drastic runtime differences for MarabouNetwork.solve() with minor query changes


SuryaB1:
Hi,

The following code for querying bound-related properties of a network to Marabou results in drastically different runtimes among seemingly minor query changes. 

Specifically, the following bound-based query script as given results in a very long runtime (>3 hours)—the execution gets stuck at the `network.solve(...)` step. For context, the network that is read sets the first output variable to 0 if the input value is <= 0, or to 1 otherwise.

However, whenever the output variable's lower bound is greater than 0.000005 as indicated in the comments below, the script executes successfully within seconds and accurately indicates that there are no satisfying values.

These runtime differences seem strange and unexpected, especially since I also noticed that output variable lower bound values smaller than 0.000005 do not cause nonterminating runtimes when I set the upper bound of the input variable to be various values above 0 (in other words, no issues for the `SAT` case). Is there some sort of precision limit at 0.00005 for the `UNSAT` query case?

Or could this be a bug? Thank you for clarifying.

```python
from maraboupy import Marabou

TF_NN_FILENAME = "saved_models/sign_classif_nn_no-softmax" # Zipped folder is attached to this post

### Read network
network = Marabou.read_tf(filename = TF_NN_FILENAME, modelType="savedModel_v2")
inputVars = network.inputVars[0][0]
outputVars = network.outputVars[0].flatten()

### Set query's input variable bounds
network.setLowerBound(inputVars[0], -999.999995)
network.setUpperBound(inputVars[0], 0.0) # finds satisfying example when upper bound is above 0.0.

### Set query's output variable bounds
lowerBound = 0.000005 # works when greater than the current value, but runs for a long time when <= 0.000005
network.setLowerBound(outputVars[0], lowerBound)

### Query Marabou
options = Marabou.createOptions(snc=False)
exitCode, vals, stats = network.solve(verbose=False, options=options)
print("vals:", vals)
```

[sign_classif_nn_no-softmax.zip](https://github.com/NeuralNetworkVerification/Marabou/files/12403671/sign_classif_nn_no-softmax.zip)

SuryaB1:
Here is another case that causes the query to not terminate in an expected time:

```python
from maraboupy import Marabou

TF_NN_FILENAME = "saved_models/sign_classif_nn_no-softmax" # Zipped folder is attached to this post

### Read network
network = Marabou.read_tf(filename = TF_NN_FILENAME, modelType="savedModel_v2")
[...]

PR #675, opened 2023-09-07
Add a upper limit to the number of rounds of preprocessing

wu-haoze:


guykatzz:
So, the changes themselves look okay, but I'm not sure this is what we want to do.
1. It is assumed that after the preproceesing, every variable has upper and lower bounds. If the network has 101 layers, in this case, some variables won't have a bound. At the very least, I would make this number ~1000.
2. What bug is this trying to solve? Since the loop stops when there is no progress in bound tightening, the only non-termination case is where a sequence of tiny improvements are made repeatedly (due to numerical imprecision). Perhaps it would be better to only consider a bound improvement of a certain minimum amount as progress.

wu-haoze:
> So, the changes themselves look okay, but I'm not sure this is what we want to do.
> 
> 1. It is assumed that after the preproceesing, every variable has upper and lower bounds. If the network has 101 layers, in this case, some variables won't have a bound. At the very least, I would make this number ~1000.
> 2. What bug is this trying to solve? Since the loop stops when there is no progress in bound tightening, the only non-termination case is where a sequence of tiny improvements are made repeatedly (due to numerical imprecision). Perhaps it would be better to only consider a bound improvement of a certain minimum amount as progress.

Hi Guy. This is to address this issue #666, which I also believe is due to the reason you mentioned.
Now that I think about it, I believe considering a bound improvement of a certain minimum amount is what we currently do (at least for the most part). For example: https://github.com/NeuralNetworkVerification/Marabou/blob/a9b17520fb2399f0c3cc5f5e3edca3418b2813c6/src/engine/Preprocessor.cpp#L403-L409

This means our preprocessor is probably already terminating, but is just making very tiny progress (e.g., barely larger than the minimum amount) each iteration in some corner cases.

I can make the number larger (10000 as you suggested) so that it's almost never hit.

MatthewDaggitt:
@wu-haoze an entry about this should probably have gone in the CHANGELOG? So that people know what has changed in the next version!

PR #676, opened 2023-09-11
Add documentation about the reindexOutputVar flag

wu-haoze:


maraboupy/Marabou.py
```diff
@@ -68,6 +68,7 @@ def read_onnx(filename, inputNames=None, outputNames=None, reindexOutputVars=Tru
         filename (str): Path to the ONNX file
         inputNames (list of str, optional): List of node names corresponding to inputs
         outputNames (list of str, optional): List of node names corresponding to outputs
+        reindexOutputVars: (bool): Reindex the variables so that the output variables are immediate after input variables.
```
tagomaru, line 71: Maybe it might be like below?

`reindexOutputVars (bool): Reindex the variables so that the output variables are immediate after input variables`

tagomaru (approved):
LGTM

PR #689, opened 2023-11-26
Produce proofs for Sign, Max, Absolute Value and Disjunction constraints

OmriIsacHUJI:


src/engine/MaxConstraint.cpp
```diff
@@ -190,18 +194,26 @@ void MaxConstraint::notifyLowerBound( unsigned variable, double value )
...
+            for ( const auto &element : _elements )
+                createElementTighteningRow( element );
+
         // TODO: optimize this. Don't need to recompute ALL possible bounds,
         // Can focus only on the newly learned bound and possible consequences.
         List<Tightening> tightenings;
         getEntailedTightenings( tightenings );
-        for ( const auto &tightening : tightenings )
+
+        if ( !proofs )
```
guykatzz, line 206: Why isn't the part below performed when proofs are on?
Also, currently we call "getEntailedTightenings" and then don't do anything with the returned information in that case, so maybe put that all under the 'if'. In the next function, too.

src/engine/MaxConstraint.cpp
```diff
@@ -311,15 +346,31 @@ void MaxConstraint::getEntailedTightenings( List<Tightening> &tightenings ) cons
             {
                 if ( !existsUpperBound( element ) ||
                      FloatUtils::gt( getUpperBound( element ), fUB ) )
-                    tightenings.append
-                        ( Tightening( element, fUB, Tightening::UB ) );
+                {
+                    if ( proofs )
+                    {
+                        ASSERT( _elementToTighteningRow[element] != NULL );
+                        _boundManager->tightenUpperBound( element, fUB, *_elementToTighteningRow[element] );
```
guykatzz, line 353: This is a significant divergence in behavior between the case when proofs are on and when they are off. This function doesn't register bound tightenings, it only computes them.

src/engine/MaxConstraint.cpp
```diff
@@ -311,15 +346,31 @@ void MaxConstraint::getEntailedTightenings( List<Tightening> &tightenings ) cons
...
         }
     }
 
     // fLB cannot be smaller than maxElementLB
     if ( FloatUtils::lt( fLB, maxElementLB ) )
-        tightenings.append( Tightening( _f, maxElementLB, Tightening::LB ) );
+    {
+        if ( proofs && maxElementForLB != _f )
+        {
+            ASSERT(_elements.exists( maxElementForLB ) && _elementToTighteningRow[maxElementForLB] != NULL );
```
guykatzz, line 368: same comment

src/engine/MaxConstraint.cpp
```diff
@@ -674,19 +725,33 @@ String MaxConstraint::serializeToString() const
...
+        _eliminatedElements.insert( variable );
+
         if ( _elementToAux.exists( variable ) )
         {
             unsigned aux = _elementToAux[variable];
             _elementToAux.erase( variable );
             _auxToElement.erase( aux );
         }
+
+        if ( _elementToTighteningRow.exists( variable ) && _elementToTighteningRow[variable] != NULL )
```
guykatzz, line 747: Should all this new code also be within an if (proofs) condition?

src/engine/MaxConstraint.cpp
```diff
@@ -712,3 +777,35 @@ bool MaxConstraint::haveOutOfBoundVariables() const
...
     return false;
 }
+
+void MaxConstraint::createElementTighteningRow( unsigned element )
+{
+
+    // Create the row only when needed and when not already create
+    if ( !_boundManager->getBoundExplainer() || _elementToTighteningRow[element] != NULL )
+        return;
+
```
guykatzz, line 787: I don't understand the logic here. Aren't tightenings performed according to existing rows, or according to contraint lemmas? Why do we need to create a row, then?

src/engine/PiecewiseLinearConstraint.h
```diff
@@ -453,9 +453,21 @@ class PiecewiseLinearConstraint : public ITableau::VariableWatcher
         return _cdInfeasibleCases;
     }
 
-    void setTableauAuxVar( unsigned var )
+    /*
+      Add a variable to the list of aux vars designated in the Tableau
+      second argument is used for MaxConstraints
+    */
+    virtual void addTableauAuxVar( unsigned tableauAuxVar, unsigned /* constraintAuxVar */ )
```
guykatzz, line 460: I'm guessing this is what you mentioned I would be commenting about.
Indeed, this is not how inheritance is usually applied - the parent can't know what MaxConstraints are.
If there's no other choice, keep it here, but make it more general: say that this value means X, and that children classes can use it if they need to.

Also, please explain to me why this parameter is needed, and why only for max.

src/engine/SignConstraint.cpp
```diff
@@ -393,17 +394,36 @@ void SignConstraint::notifyLowerBound( unsigned variable, double bound )
     if ( variable == _f && FloatUtils::gt( bound, -1 ) )
     {
         setPhaseStatus( PhaseStatus::SIGN_PHASE_POSITIVE );
+
```
guykatzz, line 397: remove one of the empty lines

src/proofs/PlcLemma.cpp
```diff
@@ -0,0 +1,113 @@
...
+                    const Vector<Vector<double>> &explanations,
+                    PiecewiseLinearFunctionType constraintType )
+    : _causingVars( causingVars )
+    , _affectedVar( affectedVar )
+    , _bound( bound )
+    , _causingVarBound( causingVarBound )
+    , _affectedVarBound( affectedVarBound )
+    , _constraintType( constraintType )
+{
+
```
guykatzz, line 31: remove empty line

src/proofs/PlcLemma.cpp
```diff
@@ -0,0 +1,113 @@
...
+            if ( _constraintType == ABSOLUTE_VALUE )
+                ASSERT( numOfExplanations == 2 || numOfExplanations == 1 );
+
+            _explanations = new double[numOfExplanations * proofSize];
+
+            for ( unsigned i = 0; i < numOfExplanations; ++i )
+                for ( unsigned j = 0; j < proofSize; ++j )
+                    _explanations[i * proofSize + j] = explanations[i][j];
+
```
guykatzz, line 66: remove empty line

src/proofs/SmtLibWriter.cpp
```diff
@@ -50,17 +180,24 @@ void SmtLibWriter::addTableauRow( const SparseUnsortedList &row, List<String> &i
         if ( FloatUtils::isZero( entry->_value ) )
             continue;
 
-        if ( counter == size - 1 )
-            break;
+        if ( counter != size - 1 )
+            assertRowLine += String( " ( + " );
+        else
+            assertRowLine += String( " " );
+
```
guykatzz, line 187: remove one of the empty lines

src/common/MString.cpp
```diff
@@ -211,8 +211,6 @@ bool String::endsWith( const String &suffix )
     return l1 >= l2 && _super.compare(l1 - l2, l2, suffix._super) == 0;
 }
 
-std::ostream &operator<<( std::ostream &stream, const String &string );
```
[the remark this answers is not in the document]
OmriIsacHUJI: It is no longer used by the proofs code, but we can keep it.

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -179,24 +191,38 @@ void AbsoluteValueConstraint::notifyUpperBound( unsigned variable, double bound
...
+        if ( proofs )
+        {
+            createPosTighteningRow();
+            createNegTighteningRow();
+        }
+
         if ( variable == _b )
         {
-            if ( bound > 0 )
+            if ( FloatUtils::gt( bound, 0 ) && ( !proofs || !phaseFixed() ) )
```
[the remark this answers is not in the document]
OmriIsacHUJI: If proofs are on and the phase is fixed, then we have a row in the tableau that corresponds exactly to this phase and we don't need to update a lemma so we can skip this part. An alternative solution would be to add a conditional, that will deduce bound without a lemma in the case the phase is fixed. What do you think?

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -179,24 +191,38 @@ void AbsoluteValueConstraint::notifyUpperBound( unsigned variable, double bound
...
             {
-                double fUpperBound = FloatUtils::max( bound, -getLowerBound( _b ) );
-                _boundManager->tightenUpperBound( _f, fUpperBound );
+                double fUpperBound = FloatUtils::max( bound, -getLowerBound( _b ) ) ;
+                if ( proofs )
+                    _boundManager->addLemmaExplanation( _f, fUpperBound, UPPER, { variable, variable }, UPPER, getType() );
```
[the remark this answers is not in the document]
OmriIsacHUJI: For the first question - addLemmaExplanation includes calling tightenUpperBound or tightenLowerBound (depending on the case).
For the second - this tightening cannot be explained using a linear combination of the variables. I think this is because the tightening uses information from both phases. It is not the case for the other tightenings.

src/engine/AbsoluteValueConstraint.cpp
```diff
@@ -729,17 +777,33 @@ String AbsoluteValueConstraint::serializeToString() const
 
 void AbsoluteValueConstraint::fixPhaseIfNeeded()
 {
+
+    if ( phaseFixed() )
+        return;
+
+    bool proofs = _boundManager && _boundManager->shouldProduceProofs();
```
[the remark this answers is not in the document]
OmriIsacHUJI: We can, but I wanted to import as least models as possible to avoid linkage problems.

src/engine/BoundManager.cpp
```diff
@@ -407,24 +407,61 @@ void BoundManager::updateBoundExplanationSparse( const SparseUnsortedList &row,
...
+        else if ( constraintType == ABSOLUTE_VALUE )
+        {
+            if ( causingVars.size() == 1 )
+            {
+                getExplanation( causingVars.front(), causingVarBound, explanation );
+                allExplanations.append( explanation );
+            }
+            else
+            {
+                // Add zero vectors to maintain consistency of explanations size
```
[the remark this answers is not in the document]
OmriIsacHUJI: Will do.
Two explanations are necessary in a lemma for a bound tightening of the sort f = max(upperBound(b), -lowerBound(b)). In this case, a proof checker should check not only that the value of f is as claimed, but also that it is the max of these two options.
For empty explanations - they are all initiated empty, and are reset to be empty whenever the corresponding bound is updated by a lemma.

As for the use of vectors - I understand, and this is one improvement we can perform. For the time being the explanations are kept in the proof object as arrays.

src/engine/MaxConstraint.cpp
OmriIsacHUJI: When calling getEntailedTightenings, we have all info necessary for updating the explanations and lemmas (for example, the element that causes a tightening for f), but we don't have it in the wrapping function. I thought it would be better to update the explanation directly inside getEntailedTightenings. and not to reproduce this info twice (in this function and in the next one).
I'm not sure I understand the suggestion. In any case, the calling getEntailedTightenings is a part of the proof production for MaxConstraint.

src/engine/MaxConstraint.cpp
OmriIsacHUJI: Agree. However, for proof production it is crucial that explanations and bounds are updated as soon as they are discovered, since otherwise we might have timing problems (as we previously had in the ReLU case).

src/engine/MaxConstraint.cpp
OmriIsacHUJI: They are. However when we update an explanation according to a tableau row, we need to find and access it. The tableau functions require either heavy computation, or would just copy the same row, so I think it is more elegant this way.
We do it too in the existing proof production for ReLU.

guykatzz (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Version 1.2.0
+
+* Added proof producing versions for Sign, Max, Absolute Value and Disjunction constraints
```
MatthewDaggitt, line 5: I think we're still on v1.1.0 actually as it hasn't been released. Maybe the `v1.1.0` title should `Next release` to avoid confusion?

PR #697, opened 2023-12-27
Split and verify an ONNX file into multiple subnets

tagomaru:
This PR is going to support a new feature to split an onnx file and verify and sequentially verify each subnet.

### DIFF
--- a/maraboupy/Marabou.py
+++ b/maraboupy/Marabou.py
@@ -3,6 +3,7 @@ Top contributors (to current version):
     - Christopher Lazarus
     - Kyle Julian
     - Andrew Wu
+    - Teruhiro Tagomori
     
 This file is part of the Marabou project.
 Copyright (c) 2017-2019 by the authors listed in the file AUTHORS
@@ -29,6 +30,7 @@ try:
     from maraboupy.MarabouNetworkONNX import *
 except ImportError:
     warnings.warn("ONNX parser is unavailable because onnx or onnxruntime packages are not installed")
+from maraboupy.MarabouNetworkComposition import MarabouNetworkComposition
 
 def read_nnet(filename, normalize=False):
     """Constructs a MarabouNetworkNnet object from a .nnet file
@@ -73,6 +75,20 @@ def read_onnx(filename, inputNames=None, outputNames=None, reindexOutputVars=Tru
         :class:`~maraboupy.MarabouNetworkONNX.MarabouNetworkONNX`
     """
     return MarabouNetworkONNX(filename, inputNames, outputNames, reindexOutputVars=reindexOutputVars)
+
+def read_onnx_with_threshould(filename, inputNames=None, outputNames=None, reindexOutputVars=True, threshold=None):
+    """Constructs a MarabouNetworkComposition object from an ONNX file
+
+    Args:
+        filename (str): Path to the ONNX file
+        inputNames (list of str, optional): List of node names corresponding to inputs
+        outputNames (list of str, optional): List of node names corresponding to outputs
+        threshold (int, optional): Threshold for the number of linear equations
+
+    Returns:
+        :class:`~maraboupy.MarabouNetworkONNX.MarabouNetworkONNX`
+    """
+    return MarabouNetworkComposition(filename, inputNames, outputNames, reindexOutputVars=reindexOutputVars, threshold=threshold)
 
 def load_query(filename):
     """Load the serialized inputQuery from the given filename
--- a/maraboupy/MarabouNetworkComposition.py
+++ b/maraboupy/MarabouNetworkComposition.py
@@ -1,0 +1,284 @@
+'''
+Top contributors (to current version):
+    - Teruhiro Tagomori
+
+This file is part of the Marabou project.
+Copyright (c) 2017-2019 by the authors listed in the file AUTHORS
+in the top-level source directory) and their institutional affiliations.
+All rights reserved. See the file COPYING in the top-level source
+directory for licensing information.
+
+MarabouNetworkComposition represents neural networks with piecewise linear constraints derived from the ONNX format
+'''
+
+import numpy as np
+import onnx
+import sys
+import pathlib
+import os
+sys.path.insert(0, os.path.join(str(pathlib.Path(__file__).parent.absolute()), "../"))
+
+from maraboupy import Marabou
+from maraboupy import MarabouCore
+from maraboupy import MarabouNetwork
+from maraboupy import MarabouNetworkONNX
+
+class MarabouNetworkComposition(MarabouNetwork.MarabouNetwork):
+    """Constructs a MarabouNetworkComposition object from an ONNX file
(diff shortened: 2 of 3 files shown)


maraboupy/Marabou.py
```diff
@@ -74,6 +76,20 @@ def read_onnx(filename, inputNames=None, outputNames=None, reindexOutputVars=Tru
     """
     return MarabouNetworkONNX(filename, inputNames, outputNames, reindexOutputVars=reindexOutputVars)
 
+def read_onnx_with_threshould(filename, inputNames=None, outputNames=None, reindexOutputVars=True, threshold=None):
```
MatthewDaggitt, line 79: Typo in first `threshold`

maraboupy/MarabouNetworkONNX.py
```diff
@@ -185,6 +190,8 @@ def processGraph(self):
         # Recursively create remaining shapes and equations as needed
         for outputName in self.outputNames:
             self.makeGraphEquations(outputName, True)
+            # if self.thresholdReached:
+            #     return
```
MatthewDaggitt, line 194: Obsolete code to be removed?

maraboupy/Marabou.py
```diff
@@ -74,6 +76,20 @@ def read_onnx(filename, inputNames=None, outputNames=None, reindexOutputVars=Tru
...
+    """Constructs a MarabouNetworkComposition object from an ONNX file
+
+    Args:
+        filename (str): Path to the ONNX file
+        inputNames (list of str, optional): List of node names corresponding to inputs
+        outputNames (list of str, optional): List of node names corresponding to outputs
+        threshold (int, optional): Threshold for the number of linear equations
```
MatthewDaggitt, line 86: A more readable name for this would be `max_number_of_lin_equations` or something similar. The current name is really not very meaningful, and the comment doesn't clear anything up. What happens when you cross the threshold?

It would also be good to provide in the docs *why* you would want to limit the number of linear equations?

maraboupy/MarabouNetworkComposition.py
```diff
@@ -0,0 +1,233 @@
...
+import os
+sys.path.insert(0, os.path.join(str(pathlib.Path(__file__).parent.absolute()), "../"))
+
+from maraboupy import Marabou
+from maraboupy import MarabouCore
+from maraboupy import MarabouNetwork
+from maraboupy import MarabouNetworkONNX
+
+class MarabouNetworkComposition(MarabouNetwork.MarabouNetwork):
+    """Constructs a MarabouNetworkComposition object from an ONNX file
```
MatthewDaggitt, line 27: This comment could be more useful. We know that the object constructs a `MarabouNetworkComposition` object. What do objects of this class actually _represent_? What do they _do_?

maraboupy/MarabouNetworkComposition.py
```diff
@@ -0,0 +1,233 @@
...
+        self.shapeMap = {}
+        self.madeGraphEquations = []
+        self.ipqs = []
+        self.ipqToInVars = {}
+        self.ipqToOutVars = {}
+        self.inputVars, self.outputVars = self.getInputOutputVars(filename, inputNames, outputNames)
+
+        network = MarabouNetworkONNX.MarabouNetworkONNX(filename, reindexOutputVars=reindexOutputVars, threshold=threshold)
+
+        network.saveQuery('q1.ipq')
```
MatthewDaggitt, line 50: Lots and lots of hard coded names here. At the very least they need to be heavily documented, but it would be better to make them configurable.

maraboupy/MarabouNetworkONNX.py
```diff
@@ -217,8 +224,15 @@ def makeGraphEquations(self, nodeName, makeEquations):
             raise RuntimeError(err_msg)
 
         # Compute node's shape and create Marabou equations as needed
+        if self.thresholdReached:
+            return
```
MatthewDaggitt, line 228: Wouldn't doing the thresholding stuff on the C++ side with the final set of equations _after_ the parsing of the network be a much better way of doing it? Then you would get this feature for all parsers and both Python and C++, rather than simply the ONNX network for the Python backend?

maraboupy/Marabou.py
tagomaru: Hi @MatthewDaggitt 
Thanks for your proactive review!
However, this PR is WIP.
So don't worry about this so much at this point. I'm working on polishing :)
Anyway, thanks again your support, I will update as you commented.

tagomaru: done

maraboupy/MarabouNetworkONNX.py
tagomaru: done

maraboupy/Marabou.py
tagomaru: done

maraboupy/MarabouNetworkComposition.py
tagomaru: done

maraboupy/MarabouNetworkComposition.py
tagomaru: reindexOutputVars was deleted since it is required

maraboupy/MarabouNetworkONNX.py
tagomaru: This is out of scope of this PR. Maybe it will be supported in the future PR.

maraboupy/Marabou.py
```diff
@@ -75,6 +77,23 @@ def read_onnx(filename, inputNames=None, outputNames=None, reindexOutputVars=Tru
...
+        outputNames (list of str, optional): List of node names corresponding to outputs
+        maxNumberOfLinearEquations (int, optional): Threshold for the number of linear equations.
+                                                    If the number of linear equations is greater than this threshold,
+                                                    the network will be split into two networks. Defaults to None.
+
+    Returns:
+        :class:`~maraboupy.MarabouNetworkComposition.MarabouNetworkComposition`
+    """
+    return MarabouNetworkComposition(filename, inputNames, outputNames,
+                                        maxNumberOfLinearEquations=maxNumberOfLinearEquations)
```
wu-haoze, line 95: indentation is off

maraboupy/MarabouNetworkONNX.py
```diff
@@ -217,8 +225,15 @@ def makeGraphEquations(self, nodeName, makeEquations):
...
         self.makeMarabouEquations(nodeName, makeEquations)
 
+        if self.maxNumberOfLinearEquations is not None:
+            if not self.thresholdReached and len(self.equList) > self.maxNumberOfLinearEquations:
+                if self.splitNetworkAtNode(nodeName, networkNamePostSplit='post_split.onnx'):
```
wu-haoze, line 234: I think one should probably use a tempfile name, otherwise two threads running in parallel would be interfering

PR #698, opened 2023-12-28
Adding support for vnnlib files encoding properties + adding support …

idan0610:
…for "Sub" operator in ONNX


guykatzz (approved):

MatthewDaggitt:
Is there a reason you've rewritten a whole parser instead of just using the existing `vnnlib` python library? See here: https://pypi.org/project/vnnlib/

MatthewDaggitt:
Also I really feel that this functionality should be added to the core C++ utility rather than the Python add on, but I recognise that that may be a losing battle. 

Finally, can you add an entry to the CHANGELOG file so that people know what has changed in the next release?

idan0610:
About using the vnnlib python module - I admit that I was not aware of it, and probably it would have really simplified the work and produced more elegant code. I will fix the implementation to use this module soon and upload the new implementation. I will also note the changes I made in the CHANGELOG file.

Regarding adding support for vnnlib through the C++ interface instead of Python - as far as I understand, the C++ interface does not currently support networks in ONNX format as input at all, so such a change is possible, but will require much more work. I think it is better to add the functionality through Python at least for the time being, and in the future consider adding it to C++ as well.

MatthewDaggitt:
> Regarding adding support for vnnlib through the C++ interface instead of Python - as far as I understand, the C++ interface does not currently support networks in ONNX format as input at all, so such a change is possible, but will require much more work.

I added support for ONNX networks to the C++ interface last year, so it should be easier than you thought :crossed_fingers: I'm just saying it because if you add the functionality to the core C++ then it can be exposed via the Python bindings easily, but if you add it to the Python it's impossible to expose via the C++ and the work has to be duplicated again.

MatthewDaggitt:
Before this does get merged in you will also need to update the `README` file on the front page as well I think...

idan0610:
Added the required note in CHANGELOG and a test for parsing a simple vnnlib. I also tried replacing my parser implementation with that of Python's vnnlib module, but it still seemed unstable and had parsing bugs even for relatively simple vnnlib files. That's why I left my implementation for the time being, and we can switch in the future to use the vnnlib module when it will be more stable.

Regarding moving the vnnlib format support to C++ - I think it's a good idea, and I have no problem starting work on it in the near future. The reason I still fixed the code for Python is because it didn't require much effort from me, and I thought it would be good to at least have the implementation in Python for now. Introducing the support for C++ is something that will probably be more complicated in any case, since I haven't had the chance to get to know the C++ code in depth so far, and it can take some time. It is my understanding that there is no library for C++ that will implement a parser for vnnlib, so I will have to implement the parser myself there as well.

Issue #641, opened 2023-06-06
Temporarily disable non-deterministic regression test failure


MatthewDaggitt:
This test seems to fail non-deterministically with about ~60% chance on CI (see e.g. https://github.com/NeuralNetworkVerification/Marabou/pull/640, and many of my other recent PRs).

This PR temporarily disables it, until the root cause can be found.

wenkokke:
This test is now disabled on master.

PR #701, opened 2024-01-06
Fix bugs in SnC

wu-haoze:
SnC becomes buggy after context-dependent objects are introduced. This problem was undetected because the regression testing system were not properly set up to test the SnC mode. After properly setting up the regression test, failed test cases related to binarized neural networks were detected.

This PR fixes the issues mentioned above.


guykatzz:
Hi Andrew,
Can you briefly explain what was the problem? There seems to be a bunch of different fixes in this PR:

1. A numerical stability issue in the explicit basis bound tightener
2. Option compatibility issues involving SOI/POI and proofs
3. Some GurobiWrapper initialization issues

and probably a few others

wu-haoze:
> Hi Andrew, Can you briefly explain what was the problem? There seems to be a bunch of different fixes in this PR:
> 
> 1. A numerical stability issue in the explicit basis bound tightener
> 2. Option compatibility issues involving SOI/POI and proofs
> 3. Some GurobiWrapper initialization issues
> 
> and probably a few others

Hi Guy, here are the fixes related to SnC:
   1) the divide and conquer parallel mode and the "portfolio" mode (each thread running a different random seed on the original problem) cannot be set to true at the same time.
   2) When a worker handles a new sub-problem (represented as a PiecewiseLinearCaseSplit), the corresponding engine needs to first pop the context level to 0 (which reverts back to the original query) and perform a new context push, before applying that PiecewiseLinearCaseSplit. We did not manipulate context at all during SnC and the effect was that previously applied case splits were never removed.
   3) The ReLU-splitting heuristics was also behaving incorrectly, in a manner similar to 2).
   4) When Gurobi is used as the LP solver,  when solving a new sub-problem, the gurobi object used for solving the previous sub-problem needs to be reset. otherwise, we would add the same equation twice.

Item 2 is the main issue. It's unfortunate how this glaring issue was overlooked. 

After also fixing the regression test system, which now actually tests the snc mode (previously, --num-workers is set to 2, but the snc mode is not on), the solver returns UNSAT on two satisfiable BNN regression benchmarks in the --snc mode. However, this is not directly related to SnC (see #641) . What I found is that explicitBasisBoundTightening derives bounds like 1.0000000001 <= x <= 0.999999999 on variables, which results the UNSAT.  when I turn off explicitBasisBoundTightening, the unsoundness issue goes away. The issue also goes away if I add these outward rounding constants.

guykatzz (approved):

PR #709, opened 2024-01-31
Fix for disjunction bug with input variables + update tests

idan0610:


src/input_parsers/VnnLibParser.cpp
```diff
@@ -101,7 +102,7 @@ int VnnLibParser::parseScript( const Vector<String> &tokens, InputQuery &inputQu
 {
     int index = 0;
 
-    while ( (unsigned) index < tokens.size() )
+    while ( ( unsigned ) index < tokens.size() )
```
guykatzz, line 105: it should actually be ( (unsigned)index < ... )

src/engine/Engine.cpp
```diff
@@ -3712,3 +3714,82 @@ void Engine::extractBounds( InputQuery &inputQuery )
         }
     }
 }
+
+void Engine::findMissingInputBoundsFromDisjunctions( InputQuery &inputQuery )
```
guykatzz, line 3718: So, this code doesn't really belong here. This is disjunction-specific code, so it belongs in the Disjunction class. I would add an abstract method to PLConstraints, called something like "return minimal LB" and "return maximal UB". Then, make a pass on the constraints, ask them for these bounds, and store them in the input query. I'm not sure whether that should be done in the Engine or in the Preprocessor.

guykatzz (commented):
This indeed looks much nicer!

One more TODO, though: can we add unit tests for the new C++ code? We should definitely have tests for the new functionality within DisjunctionContraint, and we should also add a relevant test to the Preprocessor test suite

guykatzz (approved):

PR #712, opened 2024-02-05
get rid of torch as requirement

wu-haoze:
This PR requires a slightly higher version of onnx (>=1.15) which is unsupported in python 3.7. Since python 3.8 was released as early as 2019, I think we could drop support for python 3.7 moving forward.


MatthewDaggitt:
If you're updating the ONNX on the Python side of things, you should also update ONNX on the C++ side to match: https://github.com/NeuralNetworkVerification/Marabou/blob/master/tools/download_onnx.sh.

It would also be good to have a note in the `.toml` file to that effect as well, so we don't forget in future.

Finally, this is definitely deserving of an entry in the CHANGELOG.

wu-haoze:
> If you're updating the ONNX on the Python side of things, you should also update ONNX on the C++ side to match: https://github.com/NeuralNetworkVerification/Marabou/blob/master/tools/download_onnx.sh.
> 
> It would also be good to have a note in the `.toml` file to that effect as well, so we don't forget in future.
> 
> Finally, this is definitely deserving of an entry in the CHANGELOG.

Updated download_onnx.sh, added a note in the toml file, and updated the CHANGELOG

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 * Added support for onnx networks with tanh nodes to command-line interface.
 * Added proof producing versions of Sign, Max, Absolute Value and Disjunction constraints
 * Added support for properties provided in VNN-LIB format for ONNX networks via the Python API.
-* Support additional non-linear constraints Softmax and Bilinear
+* Supported additional non-linear constraints Softmax and Bilinear
+* Removed dependency on torch and drop support for Python3.7
```
MatthewDaggitt, line 8: Should probably add a bullet point mentioning that the ONNX constraints have been bumped as well.

pyproject.toml
```diff
@@ -9,20 +9,20 @@ authors = [{ name = "The Marabou Development Team" }]
...
 license = { file = "COPYING" }
-requires-python = ">=3.7,<3.12"
+requires-python = ">=3.8,<3.12"
 
 [project.optional-dependencies]
 test = [
   # NOTE: These dependencies must be kept in-sync with the legacy requirements file 'maraboupy/test_requirements.txt'
   "numpy>=1.21.0,<2",
-  "onnx>=1.12.0,<2",
+  # NOTE: Marabou requires onnx.reference which was introduced in onnx==1.15.0
```
MatthewDaggitt, line 18: Apologies, I wasn't clear enough before. I meant that we should add a comment here saying: "The ONNX version must be kept in sync with `tools/download_onnx.sh`"

MatthewDaggitt (changes requested):

CHANGELOG.md
wu-haoze: Added the bullet point

pyproject.toml
wu-haoze: Ah I see. Added the note.

idan0610 (approved):

MatthewDaggitt (approved):

PR #715, opened 2024-02-07
Support for quantization operations

wu-haoze:
Techniques in https://arxiv.org/abs/2312.12679

- Encode Clip using two ReLUs
- Introducing Round as NonlinearConstraint


src/engine/tests/Test_MILPEncoder.h
```diff
@@ -1241,6 +1241,190 @@ class MILPEncoderTestSuite : public CxxTest::TestSuite
...
+        TS_ASSERT_THROWS_NOTHING( milp.encodeInputQuery( gurobi, inputQuery ) );
+
+        TS_ASSERT_THROWS_NOTHING( gurobi.solve() );
+
+        TS_ASSERT( gurobi.haveFeasibleSolution() );
+
+        Map<String, double> solution;
+        double costValue;
+        TS_ASSERT_THROWS_NOTHING( gurobi.extractSolution( solution, costValue ) );
+        TS_ASSERT( FloatUtils::areEqual( solution["x3"], 3.0 ) );
```
tagomaru, line 1313: Don't you need tests for cases below?

- max(x2) < 2.5
- x2 > 2.5 and x2 < 3.0
- x2 = 2.5
- x2 = 3.0

maraboupy/test/test_core.py
```diff
@@ -21,6 +21,84 @@ def test_main():
     exitCode = MarabouCore.maraboupyMain(["Marabou", "--version"])
     assert exitCode == 0
 
+def test_solve_round_and_clip_unsat():
+    """
+    -1 <= x0 <= 2
+    -2 <= x1 <= 1
+    x2 = x0 + x1
+    x3 = Clip(x2, 0.5, 1.5) (so x4 is between 0.5 and 1.5)
```
tagomaru, line 29: not x4 but x3?

maraboupy/test/test_core.py
```diff
@@ -21,6 +21,84 @@ def test_main():
...
+    2.5 <= x4
+    x4 = Round(x3) (so x5 is between 0 and 2)
```
tagomaru, line 31: not x5 but x4?

maraboupy/test/test_core.py
```diff
@@ -21,6 +21,84 @@ def test_main():
...
+
+
+def test_solve_round_and_clip_sat():
+    """
+    -1 <= x0 <= 0
+    -2 <= x1 <= 0
+    x2 = x0 + x1
+    x3 = Clip(x2, 0.5, 1.5) (x3 is fixed to 0.5)
+    -1 <= x4 <= 3
+    x4 = Round(x3) (so x3 is 0)
```
tagomaru, line 69: not x3 but x4?

src/engine/tests/Test_MILPEncoder.h
```diff
@@ -1241,6 +1241,190 @@ class MILPEncoderTestSuite : public CxxTest::TestSuite
...
+        TS_ASSERT_THROWS_NOTHING( milp.encodeInputQuery( gurobi, inputQuery ) );
+
+        TS_ASSERT_THROWS_NOTHING( gurobi.solve() );
+
+        TS_ASSERT( gurobi.haveFeasibleSolution() );
+
+        Map<String, double> solution;
+        double costValue;
+        TS_ASSERT_THROWS_NOTHING( gurobi.extractSolution( solution, costValue ) );
+        TS_ASSERT( FloatUtils::areEqual( solution["x1"], 2 ) );
```
tagomaru, line 1425: Don't you need the case of equidistant from two integers, e.g. 1.5 <= x0 <= 2.5?

maraboupy/test/test_core.py
wu-haoze: Fixed

maraboupy/test/test_core.py
wu-haoze: fixed

maraboupy/test/test_core.py
wu-haoze: fixed

src/engine/tests/Test_MILPEncoder.h
wu-haoze: Added test for this case.

src/engine/tests/Test_MILPEncoder.h
wu-haoze: Added test for these cases

src/engine/tests/Test_MILPEncoder.h
tagomaru: I checked the added the test "test_encode_round_constraint4" and the assertion below.

```
        TS_ASSERT( FloatUtils::areEqual( solution["x1"], 1 ) ||
                   FloatUtils::areEqual( solution["x1"], 2 ) ||
                   FloatUtils::areEqual( solution["x1"], 3 ) );
```

I thought this expectation of the first line meant x1 was 1.0 when x0 = 1.
If it's correct, you might have to add `FloatUtils::areEqual( solution["x1"], 1) ||` as well in test_encode_round_constraint1?

wu-haoze: In this test, we have x0 between 1.5 and 2.5. But since our encoding does **not** precisely encode the equidistant case, we should expect the MILP solver to return anything between 1, 2, and 3. 

There is no such ambiguity in test_encode_round_constraint1 though, because x0 is between 1.2 and 1.3, and x1 can only be equal to 1.

wu-haoze: The reason we can distinguish all cases but equidistant case is because in MILP we encode f = Round(b) as the following:     

f - b <= 0.5
b - f <= 0.5

where f is an integer.

tagomaru (approved):
LGTM

src/common/FloatUtils.h
huangdiudiu, line 1: I'm not sure if it would be better to define f and c as INT type here

src/common/FloatUtils.h
```diff
@@ -106,6 +106,20 @@ class FloatUtils
         return gt( x, y, epsilon ) ? x : y;
     }
 
+    static double round( double x, double epsilon = GlobalConfiguration::DEFAULT_EPSILON_FOR_COMPARISONS )
+    {
+        // # Implement pytorch round https://pytorch.org/docs/stable/generated/torch.round.html
+        // "Values equidistant from two integers are rounded towards the nearest even value"
+        // i.e., round(2.5) == round(1.5) == 2
+        double f = floor( x );
```
huangdiudiu, line 114: I'm not sure if it would be better to define f and c as INT type here

src/engine/InputQuery.cpp
```diff
@@ -157,6 +163,54 @@ void InputQuery::addPiecewiseLinearConstraint( PiecewiseLinearConstraint *constr
     _plConstraints.append( constraint );
 }
 
+void InputQuery::addClipConstraint( unsigned b, unsigned f,
```
huangdiudiu, line 166: Do we need to support infinite upper and lower bounds?  like floor=-infinity or ceiling=+infinity？

wu-haoze:
You mean like a Clip with only floor or only ceiling? I agree there is a motivation for this. Could you create an issue on this?

________________________________

PR #716, opened 2024-02-07
remove redundant addend in equation

wu-haoze:
This PR removes redundant addends in the equation before invoking preprocessing. Redundant means there are two addends that have the same variable, or there is an added with 0 coefficient.


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Bumped ONNX version to >= 1.15.0
 * Added support for Leaky ReLU
 * Added support for Pythonic API
+* Remove redundant addends in Equation objects
```
MatthewDaggitt, line 12: So I don't think this should go in the CHANGELOG because it's not an observable change to the user?

src/engine/Equation.h
```diff
@@ -86,6 +86,22 @@ class Equation
     */
     void setCoefficient( unsigned variable, double newCoefficient );
 
+    /*
+      Transform the equation so that any two addends have different variables
+      and no addends have zero coefficients
+
+      CAUTION: Should only be used during preprocessing.
```
MatthewDaggitt, line 93: Adding _why_ it should only be used during preprocessing would be helpful.

CHANGELOG.md
wu-haoze: I see. Sure. I removed it from the CHANGELOG

src/engine/Equation.h
wu-haoze: Clarified

MatthewDaggitt (approved):

PR #725, opened 2024-02-10
Print errors on stderr rather than stdout

### LINKED ISSUE #714: Marabou outputs error messages on `stdout` rather than `stderr`
Currently when Marabou errors, it outputs the messages to the standard output stream instead of the standard error stream. This makes interacting it with external programs a little tricky.

I haven't yet investigated when and where it's doing this.

MatthewDaggitt:
Fixes #714. Tested locally.


src/engine/MarabouMain.cpp
```diff
@@ -128,7 +128,7 @@ int marabouMain( int argc, char **argv )
     }
     catch ( const Error &e )
     {
-        printf( "Caught a %s error. Code: %u, Errno: %i, Message: %s.\n",
+        fprintf(stderr, "Caught a %s error. Code: %u, Errno: %i, Message: %s.\n",
```
wu-haoze, line 131: frptinf(stderr -> fprintf( stderr

This is the coding convention of files in src.

wu-haoze:
The test fail because of a hole in the snc logic. Should pass after applying this patch: #726

src/engine/MarabouMain.cpp
MatthewDaggitt: Fixed :+1:

wu-haoze (approved):

PR #731, opened 2024-02-12
Support multiple outputs in C++ ONNX parser

MatthewDaggitt:


src/input_parsers/OnnxParser.cpp
```diff
@@ -65,8 +65,8 @@ OnnxParser::OnnxParser( const String &path )
 void OnnxParser::generateQuery( InputQuery& query )
 {
     Set<String> inputNames = readInputNames();
-    String outputName = readOutputName();
-    processGraph( inputNames, outputName, query );
+    Set<String> terminalNames = readOutputNames();
```
wu-haoze, line 68: Why use terminal names instead of output names?

src/input_parsers/OnnxParser.cpp
```diff
@@ -75,14 +75,14 @@ void OnnxParser::generateQuery( InputQuery& query )
  * @param query The query object to be populated
  * @param inputNames The set of input nodes. Note these don't have to be an actual input to the
  * network, they can be intermediate nodes.
- * @param outputName The output node. Note that again this doesn't have to be an actual output of
- * the network, it can be an intermediate node.
+ * @param outputName The set of terminal nodes. Note that these doesn't have to be outputs of
```
wu-haoze, line 78: Should be @param terminalName**s** ?

But I would prefer outputNames

src/input_parsers/OnnxParser.cpp
MatthewDaggitt: I was trying to distinguish between the output nodes of the neural of the network, and the nodes that the query terminates at (that can be arbitrary nodes in the network). If you prefer the overloaded term, then I'm happy to switch back?

wu-haoze: Ah I see. In that case I think terminal names make sense. Could you please document this distinction between actual output and specified terminals somewhere?

wu-haoze: Other than this, the PR LGTM

wu-haoze (approved):

PR #732, opened 2024-02-12
Implement the Cast operator in C++ ONNX parser

MatthewDaggitt:
The old layer zoo ONNX files have changed as we switched from v12 to v15


src/input_parsers/tests/Test_OnnxParser.h
```diff
@@ -313,4 +319,17 @@ class OnnxParserTestSuite : public CxxTest::TestSuite
...
+    {
+        Vector<double> input = {
+            0.0, 0.0,
+            0.0, 0.0
+        };
+        Vector<double> output = {
+            0.0, 1.0,
+            -1.0, 2.0
+        };
+        run_test("cast_int_to_float", input, output);
```
wu-haoze, line 333: Need space in parenthesis.

wu-haoze (approved):

PR #737, opened 2024-02-13
Clang format

wu-haoze:
1. add a workflow to perform clang format check of all .cpp and .h files according to .clang-format at the repo root
2. clang format all C++ files in src/
3. add a script tools/clang-format-src.h


src/basis_factorization/ForrestTomlinFactorization.cpp
```diff
@@ -13,15 +13,18 @@
 
  **/
 
+#include "ForrestTomlinFactorization.h"
+
 #include "BasisFactorizationError.h"
 #include "EtaMatrix.h"
 #include "FloatUtils.h"
-#include "ForrestTomlinFactorization.h"
```
guykatzz, line 19: This one is a bit strange. Why did it single out this include?

guykatzz (approved):

src/basis_factorization/ForrestTomlinFactorization.cpp
wu-haoze: I think it's trying to single out the header file of this particular cpp file. Let me try to turn this off.

wu-haoze: Not turning this off per off-line discussion.

PR #741, opened 2024-02-13
Implement LeakyRelu operation in C++ ONNX parser

MatthewDaggitt:


src/input_parsers/NetworkParser.cpp
```diff
@@ -110,6 +115,13 @@ void NetworkParser::getMarabouQuery( InputQuery& query )
         query.addPiecewiseLinearConstraint( constraintPtr );
     }
 
+    for ( LeakyReluConstraint* constraintPtr : _leakyReluList )
+    {
+        LeakyReluConstraint constraint = *constraintPtr;
```
wu-haoze, line 120: It seems we are making this call of the copy constructor just for debugging. If so, we should guard this line and the assertion below with the Debug macro so that in production mode they are ignored. See the following for an example.
https://github.com/NeuralNetworkVerification/Marabou/blob/7e2dd2b308933d436f429ec16690b5bd5513ea74/src/engine/Engine.cpp#L633-L660

You would need to include "Debug.h" 

It seems the copy constructor is invoked for some other constraints. Could you fix them as well (either here or in a separate PR)?

wu-haoze (changes requested):

src/input_parsers/NetworkParser.cpp
MatthewDaggitt: Fixed them in this PR.

PR #742, opened 2024-02-13
use alpha in leaky relu python parser

wu-haoze:


MatthewDaggitt (approved):
Probably needs a CHANGELOG entry saying a (serious?) bug was fixed in the implementation of LeakyRelU layers? (Can you add one saying several serious bugs were fixed in the command line ONNX parser as well?) 

Other than that looks good!

wu-haoze:
> Probably needs a CHANGELOG entry saying a (serious?) bug was fixed in the implementation of LeakyRelU layers? (Can you add one saying several serious bugs were fixed in the command line ONNX parser as well?)
> 
> Other than that looks good!

Since the leaky relu constraint was introduced after the initial release, I suppose this would not belong to the CHANGELOG. I agree a line could be added about bug fixes in the ONNX parser, but perhaps in a different PR?

MatthewDaggitt:
> Since the leaky relu constraint was introduced after the initial release, I suppose this would not belong to the CHANGELOG.

Ah yup! Nice catch. Didn't realise it was that new!

Issue #723, opened 2024-02-09
Using Convolutional Layer leads to 'std::out_of_range' with onnx input


progirep:
So we are currently trying to get a workflow involving neural network verification for networks that involve convolutional layers to get to run. Unfortunately, we are having little luck so far, with varying error messages. I'd like to report a 'std::out_of_range' error that can occur for some inputs.

Attached is an example Jupyter notebook for learning and writing an ONNX-file, the resulting the ONNX file, and a property file.

Running `/path/to/Marabou simple_cnn_model.onnx example_property.txt ` yields:

```
Network: simple_cnn_model.onnx
Property: example_property.txt

Engine::processInputQuery: Input query (before preprocessing): 7 equations, 14 variables
Engine::processInputQuery: Input query (after preprocessing): 0 equations, 2 variables

Input bounds:
	x0: [  4.0000,   4.0000] [FIXED]

terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
```
[example.zip](https://github.com/NeuralNetworkVerification/Marabou/files/14221929/example.zip)

Note that this the attached problem is not one that we are actually interested in, but resulted from simplifying a non-working case as much as we can. The used Marabou version is the latest version on GIT (main branch) at the time of writing.

wu-haoze:
@progirep It seems the issue is that the network you have here uses conv1d. but Marabou currently only supports conv2d.  Is it possible to extend the onnx C++ parser  to support the network in this issue?

BlackTentacle:
Thanks for the feedback! Well, if that's not supported, then it's not supported. It would be great, however, to get an error message whenever a layer type is not supported, because that would save a lot of time.

progirep:
Thanks for the feedback! Well, if that's not supported, then it's not supported. It would be great, however, to get an error message whenever a layer type is not supported, because that would save a lot of time.

MatthewDaggitt:
Hmm it really should be throwing an error:

https://github.com/NeuralNetworkVerification/Marabou/blob/f686c6bd99db76ffc3c148001e47d05095d2e8a3/src/input_parsers/OnnxParser.cpp#L801-L808

I wonder what's going on...

MatthewDaggitt:
Hmm when I run the example for myself I get the expected error:
```bash
Caught a MarabouError error. Code: 105, Errno: 0, Message: Onnx operation Unsqueeze not currently supported by Marabou.
```
from the command
```bash
[...]

PR #743, opened 2024-02-13
Two fixes addressing issue 723

progirep:
This is supposed to address issue 723: https://github.com/NeuralNetworkVerification/Marabou/issues/723

There are two changes:
- Using a non-two-D Convolution Layer now gives an error message
- If the layer sizes are stored in INT64 instead of INT32, this doesn't lead to memory access violations. The previous behavior was quite weird because when interpreting an INT64 array as an INT32 array, the "-1" at the beginning of the array (in layer sizes) get duplicated, which leads to the reshaping operation code failing in an unexpected way.

The current code is not clean in that probably the error message formatting is not exactly how this is envisioned in the project, the code formatting is not great, and data types are not checked uniformly. The additional checks for data types probably make the overall code a tiny bit slower, but not much.

### DIFF
--- a/src/input_parsers/OnnxParser.cpp
+++ b/src/input_parsers/OnnxParser.cpp
@@ -411,12 +411,26 @@ Vector<int> getTensorIntValues( const onnx::TensorProto& tensor, const TensorSha
     {
         checkEndianness();
         const char* bytes = raw_data.c_str();
-        const int* ints = reinterpret_cast<const int*>( bytes );
-        for ( int i = 0; i < size; i++ )
-        {
-            int value = *(ints + i);
-            result.append( value );
-        }
+        onnx::TensorProto_DataType dataType = static_cast<onnx::TensorProto_DataType>( tensor.data_type() );
+        
+        if (dataType == onnx::TensorProto_DataType_INT32) {
+    		const int32_t* ints = reinterpret_cast<const int32_t*>( bytes );
+		for ( int i = 0; i < size; i++ )
+		{
+		    int value = *(ints + i);
+		    result.append( value );
+		}
+	} else if (dataType == onnx::TensorProto_DataType_INT64) {
+		const int64_t* ints = reinterpret_cast<const int64_t*>( bytes );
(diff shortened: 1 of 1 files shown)


src/input_parsers/OnnxParser.cpp
```diff
@@ -411,12 +411,26 @@ Vector<int> getTensorIntValues( const onnx::TensorProto& tensor, const TensorSha
...
+		const int64_t* ints = reinterpret_cast<const int64_t*>( bytes );
+		for ( int i = 0; i < size; i++ )
+		{
+		    int value = *(ints + i);
+		    result.append( value );
+		}
+	} else {
+		String errorMessage = Stringf( "Illegal data type for integer tensors used in the model. Only INT32 and INT64 supported." );
+            throw MarabouError( MarabouError::ONNX_PARSER_ERROR, errorMessage.ascii() );
+	}
```
MatthewDaggitt, line 433: I think you also need to add a corresponding case for the `else` below when the data is not stored in `raw_data`. I'd advice lifting the `dataType` check above the `raw_data` check.

MatthewDaggitt (changes requested):
Thanks for the PR! Generally looks great, but a few comments.

wu-haoze:
@progirep  Thank you so much for your code contribution. I have two requests (sorry for the trouble)!

1. Regarding the code style, we do have one that we try to follow and are in the process of strictly enforcing it (#737). Please feel free to use the [.clang-format file](https://github.com/wu-haoze/Marabou/edit/clang-format/.clang-format?pr=%2FNeuralNetworkVerification%2FMarabou%2Fpull%2F737) to format your PR. 

2. Could you please also squash the commits into a single commit, and commit with the --signoff flag on (i.e., `git commit --signoff`). This is to affirm that your commit complies with the licensing requirement governing this repo. You might need to stash your changes, recommit, and `git push --force`. 

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository#about-commit-signoffs

PR #746, opened 2024-02-14
Do not reindex output variables in python ONNX parser by default 

wu-haoze:
Also fix a minor bug exposed by the test.


MatthewDaggitt (approved):
In the medium term I would argue that if we no longer think there's a purpose to this argument we should really remove it. There's a lot of ugly complicated code involved in supporting it. 

Whether that's deprecating it and then removing it, or taking advantage of the big v2.0 release just to remove it immediately I'm not sure. If we do remove it, we should wait until https://github.com/NeuralNetworkVerification/Marabou/pull/745 gets merged in, to avoid the merge conflicts.

PR #752, opened 2024-02-16
Touch up README.md and run `clang-format` via precommit hooks

MatthewDaggitt:
This touches up the docs and adds instructions to them on how to get `clang-format` installed as a pre-commit hook in a slightly more robust way. In particular, we use `pre-commit` to ensure there's no need for a special bash script any more, and to make it easy for everyone to ensure that formatting runs whenever they commit.

@wu-haoze if you run then everything will be setup for you.
```bash
pip install pre-commit
pre-commit install
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,16 @@
+# pre-commit configuration file
+#
+# TL;DR
+#   pre-commit runs a series of formatters and linters on every commit.
+#   If throws an error, it's usually already fixed the problem and just
+#   wants you to have a look at the changes it made.
+#   To setup the pre-commit hooks, you'll need pre-commit, and install the
+#   hooks by running `pre-commit install` from the repository root.
+#
+# See: https://pre-commit.com
+
+repos:
+- repo: https://github.com/pre-commit/mirrors-clang-format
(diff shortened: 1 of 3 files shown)


wu-haoze:
@MatthewDaggitt  Thanks for making this change!  I suggest we keep the bash script as a low commitment option for one-time developers.

wu-haoze (approved):

MatthewDaggitt:
> I suggest we keep the bash script as a low commitment option for one-time developers.

The problem is the bash script doesn't actually work for me.... In particular the shell I use doesn't calculate the path correctly with `SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"`.

I would argue that running the following once:
```
pip install pre-commit
pre-commit install
```
is less of a burden for one-time developers than running
```
apt install clang-format
```
and then much later, separately and repeatedly remembering to run
```
./tools/clang-format
```
(if the latter even works for them).

In particular I'm not sure we should be targeting "one-time" developers. We should be encouraging people to be at least "two-time" or "three-time" developers at least :wink: 

Having said that, if you would like to keep the bash script around then we can?

PR #756, opened 2024-02-17
avoid implicit type casting in oonx parser

wu-haoze:


MatthewDaggitt:
See comment in https://github.com/NeuralNetworkVerification/Marabou/pull/757 for the correct fix.

Issue #755, opened 2024-02-17
Compilation error with unsqueeze function


wu-haoze:
Pulling the latest master resulting and compiling locally with G++ 11.4.0 resulting in the following error:

```
[  1%] Building CXX object CMakeFiles/MarabouHelper.dir/src/input_parsers/OnnxParser.cpp.o
/home/haozewu/Projects/Marabou_dev/Marabou/src/input_parsers/OnnxParser.cpp: In member function ‘void OnnxParser::unsqueeze(onnx::NodeProto&)’:                                                                   
/home/haozewu/Projects/Marabou_dev/Marabou/src/input_parsers/OnnxParser.cpp:1208:66: error: conversion from ‘Vector<long int>’ to non-scalar type ‘Vector<int>’ requested                                         
 1208 |     Vector<SignedTensorIndex> axes = _constantIntTensors[axisName];
      |                                                                  ^
make[2]: *** [CMakeFiles/MarabouHelper.dir/build.make:1518: CMakeFiles/MarabouHelper.dir/src/input_parsers/OnnxParser.cpp.o] Error 1                                                                              
make[1]: *** [CMakeFiles/Makefile2:606: CMakeFiles/MarabouHelper.dir/all] Error 2
make: *** [Makefile:101: all] Error 2
```

wu-haoze:
Is #756  a reasonable fix?

MatthewDaggitt:
Yes, https://github.com/NeuralNetworkVerification/Marabou/pull/753 and https://github.com/NeuralNetworkVerification/Marabou/pull/748 had a semantic merge conflict. By default Github only picks up on syntactic merge conflicts. To avoid this happening in future, we should turn on Github's merge queue feature!

PR #757, opened 2024-02-17
Refactor C++ ONNXParser to a pure style that takes in an InputQueryBuilder as an argument.

MatthewDaggitt:
Now matches the corresponding Python files almost perfectly.

### DIFF
--- a/src/input_parsers/OnnxParser.cpp
+++ b/src/input_parsers/OnnxParser.cpp
@@ -40,54 +40,27 @@
  * Public methods *
  ******************/
 
-OnnxParser::OnnxParser( const String &path )
-{
-    // open file and move current position in file to the end
-    std::ifstream input( path.ascii(), std::ios::ate | std::ios::binary );
-    // get current position in file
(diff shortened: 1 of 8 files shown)


src/input_parsers/OnnxParser.cpp
```diff
@@ -1205,14 +1217,15 @@ void OnnxParser::unsqueeze( onnx::NodeProto &node )
     {
         missingNodeError( axisName );
     }
-    Vector<SignedTensorIndex> axes = _constantIntTensors[axisName];
+    Vector<int64_t> axes = _constantIntTensors[axisName];
```
MatthewDaggitt, line 1220: This and the changes from 1225-1228 are the correct fix for #755. We can merge this in, or I can port the fix across to #756 if you'd like @wu-haoze ?

wu-haoze: Hi Matthew, merge this fix in here sounds good.

wu-haoze (approved):

PR #760, opened 2024-02-17
Backward analysis

### LINKED ISSUE #730: Assertion violation error when using MILP with --milp-tightening=milp in Debug mode
When Using Marabou in debug mode with MILP (Gurobi academic licence) and running relatively simple queries, like the ONNX network at resources/onnx/vnnlib/test_sat_vnncomp.onnx and the property file resources/onnx/vnnlib/test_prop_vnncomp.vnnlib, with the following command (from the build directory):

```
../resources/onnx/vnnlib/test_sat_vnncomp.onnx ../resources/onnx/vnnlib/test_prop_vnncomp.vnnlib --milp --milp-tightening milp
```

I get the assertion error:

```
Assertion violation! File /home/idan0610/Marabou_dev/Marabou/src/common/GurobiWrapper.cpp, line 275
```

When running this same query in Release mode, a segfault error is received.
Lines 275-276 in GurobiWrapper.cpp (function GurobiWeapper::setCost):

```
ASSERT( _nameToVariable.exists( term._variable ) );
cost += GRBLinExpr( *_nameToVariable[term._variable], term._coefficient );
```

from what I could tell, setCost() is called from NLR::MILPFormulator::optimzeWithGurobi(), which in turn is called from NLR::MILPFormulator::tightenSingleVariableBoundsWithMILPEncoding() in line 507. A few lines before that, in line 492, there is a call to gurobi->reset(), which empties field map _nameToVariable, and no items are added to this map after that, causing the error.

wu-haoze:
- Introduces an additional mode to perform backward LP-tightening as described in https://arxiv.org/abs/2203.03153 

- Also fix #730

### DIFF
--- a/src/common/GurobiWrapper.cpp
+++ b/src/common/GurobiWrapper.cpp
@@ -93,7 +93,6 @@ void GurobiWrapper::resetModel()
 
 void GurobiWrapper::reset()
 {
-    _nameToVariable.clear();
     _model->reset();
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,17 @@
 
 ## Next Release
 
+* Changes in core solving module:
+  - Added proof producing versions of `Sign`, `Max`, `Absolute Value` and `Disjunction` constraints.
+  - Added support for `LeakyRelu`, `Clip`, `Round`, `Softmax`.
+  - Added support for forward-backward abstract interpretation.
+
 * Dependency changes:
   - Dropped support for Python 3.7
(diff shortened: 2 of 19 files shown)


src/common/GurobiWrapper.cpp
```diff
@@ -93,7 +93,6 @@ void GurobiWrapper::resetModel()
 
 void GurobiWrapper::reset()
 {
-    _nameToVariable.clear();
     _model->reset();
```
wu-haoze, line 96: @idan0610 , the fix is to simply not clear this map when calling reset(), which is intended to reset the solver without removing the constraints.

idan0610 (approved):

PR #767, opened 2024-02-21
Refactor CMakeList and download scripts to use versions from a single source

MatthewDaggitt:
Versions are all set once in the CMakeLists file


wu-haoze (approved):
Hi Matthew, thanks for the

PR #768, opened 2024-02-26
Fix python API examples

wu-haoze:
There are some path and index issues with the python API examples. This PR fixes them. Now all the examples terminate fine.

### DIFF
--- a/maraboupy/examples/0_NNetExample.py
+++ b/maraboupy/examples/0_NNetExample.py
@@ -30,7 +30,7 @@ nnetFile = "../../src/input_parsers/acas_example/ACASXU_run2a_1_1_tiny_2.nnet"
 # %%
 # Load the network from NNet file, and set a lower bound on first output variable
 net1 = Marabou.read_nnet(nnetFile)
-net1.setLowerBound(net1.outputVars[0][0], .5)
+net1.setLowerBound(net1.outputVars[0][0][0], .5)
 
 # %%
 # Solve Marabou query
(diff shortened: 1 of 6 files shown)


maraboupy/examples/0_NNetExample.py
```diff
@@ -5,21 +5,22 @@
...
+import pathlib
 import sys
 import numpy as np
 
 ## %
-# Path to Marabou folder if you did not export it
-
-# sys.path.append('/home/USER/git/Marabou')
+# Add path to Marabou folder if you did not export it
+sys.path.insert(0, os.path.join(str(pathlib.Path(__file__).parent.absolute()), "../../"))
```
MatthewDaggitt, line 23: While I understand the motivation for this, I personally feel this may be setting users up for more pain. Are we going to get a lot of people saying "the examples file works just fine, but why doesn't my own file work?"? 

My own preference would be to have people getting a working installation sooner rather than later. Having said that, don't feel too strongly about it, so if you think is the best way happy to defer.

MatthewDaggitt (changes requested):
Can we add running these files to the Python test suites on CI? So we don't have more similar regressions in future?

maraboupy/examples/0_NNetExample.py
wu-haoze: This makes sense. I undid changes to the path but only fix the index issue.

wu-haoze:
> Can we add running these files to the Python test suites on CI? So we don't have similar regressions in future?

Let me add a workflow to run these files.

MatthewDaggitt (approved):

PR #770, opened 2024-02-27
C++11 -> C++17

wu-haoze:


CMakeLists.txt
```diff
@@ -80,6 +80,7 @@ include_directories(SYSTEM ${CVC4_DIR} ${CVC4_DIR}/include)
 ## Boost ##
 ###########
 
+add_definitions(-DBOOST_NO_CXX98_FUNCTION_BASE)
```
wu-haoze, line 83: Fix described in https://github.com/boostorg/container_hash/issues/22

MatthewDaggitt: Might be best to add this as a comment to the file itself. Otherwise this explanation is likely to get lost?

MatthewDaggitt (approved):
Well that was easy :crossed_fingers: I'll try rebasing my Pytorch branch on this.

CMakeLists.txt
wu-haoze: Not sure this comment will be too relevant later on?

MatthewDaggitt: I think if someone is coming in three years time and is wondering why this line exists and whether it's still needed, then it would be useful documentation?

wu-haoze: They could probably just look this flag up. :) But sure, doesn't hurt to add the comment.

idan0610 (approved):

PR #787, opened 2024-04-02
Fix in input splitting logic

wu-haoze:


src/engine/Engine.cpp
```diff
@@ -2838,10 +2838,14 @@ PiecewiseLinearConstraint *Engine::pickSplitPLConstraint( DivideStrategy strateg
         candidatePLConstraint = pickSplitPLConstraintBasedOnPolarity();
     else if ( strategy == DivideStrategy::EarliestReLU )
         candidatePLConstraint = pickSplitPLConstraintBasedOnTopology();
-    else if ( strategy == DivideStrategy::LargestInterval &&
-              ( _smtCore.getStackDepth() % GlobalConfiguration::INTERVAL_SPLITTING_FREQUENCY ==
+    else if ( !_produceUNSATProofs && strategy == DivideStrategy::LargestInterval &&
```
omriisack, line 2841: If we turn this option off when producing proofs (as I agree we should), shouldn't we add !_produceUNSATProofs as a condition for choosing DivideStrategy::LargestInterval in the function Engine::decideBranchingHeuristics as well?

wu-haoze: That's a good point. Let me do that.

wu-haoze: UPDATE: actually, I should only turn input splitting off in decideBranchingHeuristics(), instead of during the actual solving.

omriisack (approved):

PR #794, opened 2024-04-08
Update pyproject.toml

wu-haoze:


wenkokke:
Just run `pipx run bumpver update --major --dry`, review the diff, and then repeat the command without the `--dry`, instead of this PR.

PR #796, opened 2024-04-09
bump  supported python versions in pyproject.toml

wu-haoze:
Will create a PR to bump the version to 2.0 after this is merged.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -62,7 +62,7 @@ jobs:
         env:
           MACOSX_DEPLOYMENT_TARGET: "10.9"
 
-      - uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


wenkokke:
You need to bump the version of the cibuildwheel action in the cibuildwheel workflow to include Python 3.12 builds.

wu-haoze:
> You need to bump the version of the cibuildwheel action in the cibuildwheel workflow to include Python 3.12 builds.

You mean like this?
 https://github.com/NeuralNetworkVerification/Marabou/pull/796/commits/2ba47cd4996581da9762b18385b577593c3fb5a5

wenkokke:
No, the action called `cibuildwheel` should be bumped to the latest version, which I think is `2.17`?

wenkokke (approved):

PR #797, opened 2024-04-10
bump version v1.0.0 -> v2.0.0

wu-haoze:


wenkokke (approved):

MatthewDaggitt:
It might be good to write down the release procedure in the developer docs for the future?

wu-haoze:
> It might be good to write down the release procedure in the developer docs for the future?

Will do after getting the procedure to work.

PR #799, opened 2024-04-11
Fix github action failure to release

wu-haoze:


wenkokke:
The `wheelhouse-*-*` patterns are there to match the uploads, so you need to make sure you upload the wheels with a unique name for each job, then match those names here.

wenkokke:
This should be v4, because upload and download v3 and v4 are not mutually compatible:

```yaml
# cibuuldwheel.yml
- uses: actions/upload-artifact@v3
        with:
          name: wheelhouse
          path: ./wheelhouse/*.whl
```

The name field should be some identifier that is unique for each run of that job. Probably `wheelhouse-${{ runner.os }}` suffices.

The pattern in the download should then match that, so `wheelhouse-*` in that case.

wenkokke (changes requested):
See comments.

wenkokke:
You'll have to recreate the tag.

PR #831, opened 2024-08-22
Incremental solving infrastructure

wu-haoze:
1. Old InputQuery class renamed to "Query"
2. New InputQuery class introduced which allows pushing/popping constraints incrementally. Keep naming the outward-facing query class InputQuery makes the API backward compatible.
3. InputQuery and Query both inherits an IQuery abstract class. QueryLoaders, different parsers, and methods like preprocessQuery now operate on IQuery. 
4. Update incremental-solving example in maraboupy/example


src/engine/DnCManager.cpp
```diff
@@ -347,39 +329,23 @@ void DnCManager::printResult()
     {
         std::cout << "sat\n" << std::endl;
 
-        extractSolution( *_baseInputQuery );
-
-        Vector<double> inputVector( _baseInputQuery->getNumInputVariables() );
```
wu-haoze, line 352: There is a danger that the network-level reasoner's output variable is different from the actual output variable. Also, the output assignment might be inconsistent with the satisfying assignment returned by the API call. Disabling this re-computation with NLR for now and left a TODO.

src/engine/Engine.cpp
```diff
@@ -918,19 +918,18 @@ void Engine::fixViolatedPlConstraintIfPossible()
     _tableau->setNonBasicAssignment( fix._variable, fix._value, true );
 }
 
-bool Engine::processInputQuery( InputQuery &inputQuery )
+bool Engine::processInputQuery( const IQuery &inputQuery )
```
wu-haoze, line 921: Making this const so that we know processInputQuery does not change the InputQuery, which will be "re-processed" later during incremental solving.

src/engine/Engine.cpp
```diff
@@ -973,35 +972,12 @@ bool Engine::calculateBounds( InputQuery &inputQuery )
     return true;
 }
 
-void Engine::informConstraintsOfInitialBounds( InputQuery &inputQuery ) const
```
wu-haoze, line 976: Moving this to Preprocessor. This is the first thing we do now in the Preprocessor.

maraboupy/MarabouCore.cpp
```diff
@@ -376,42 +376,6 @@ struct MarabouOptions
 };
 
 
-/* The default parameters here are just for readability, you should specify
```
wu-haoze, line 379: I believe this method is not needed given that we already have calculuateBounds(). Writing a deep-copy constructor for the context dependent InputQuery class just for this seems a bit unncessary.

src/input_parsers/PropertyParser.h
```diff
@@ -17,31 +17,28 @@
...
 #include "Equation.h"
-#include "InputQuery.h"
+#include "IQuery.h"
 #include "MString.h"
+#include "Query.h"
 
 /*
   This class reads a property from a text file, and stores the property's
-  constraints within an InputQuery object.
+  constraints within an Query object.
```
MatthewDaggitt, line 26: But the signature takes an `IQuery`, not a `Query`?

MatthewDaggitt (changes requested):
Otherwise, the bits that I understand look good! Obviously can't comment on the alterations to the preprocessing.

deps/CVC4/context/cdhashmap.h
```diff
@@ -449,6 +449,21 @@ class CDHashMap : public ContextObj {
     }
   }
 
+  // The following two methods were added for the Marabou project
+  Data get( const Key &k ) const
```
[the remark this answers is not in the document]
wu-haoze: Yeah, I agree it's a bit sloppy here to directly edit dependency code. Perhaps I could define a class that inherits cdhashmap and contains these two additional methods. 

I believe we are using the correct data structure though. I added unit tests for this data structure to be sure. get() and exists() are methods that data structures in this repo tends to have. In CVC4, they would use the find() method in the cdhashmap class.

maraboupy/examples/6_IncrementalSolvingExample.py
```diff
@@ -44,8 +45,10 @@
 # The set of images to check local robustness against
 random_images = [np.random.random((784,1)) for _ in range(NUM_SAMPLES)]
 
-# Step 1: load the network
+# Step 1: load the network and get the input query
 network = Marabou.read_onnx(filename)
+ipq = network.getInputQuery()
```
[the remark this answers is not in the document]
wu-haoze: I agree. :) Let me rename them all to query in the examples.

src/engine/InputQuery.h
```diff
@@ -16,51 +16,95 @@
...
 #include "NonlinearConstraint.h"
 #include "PiecewiseLinearConstraint.h"
+#include "context/cdo.h"
 
-class InputQuery
+#include <context/cdhashmap.h>
+#include <context/cdlist.h>
+#include <context/context.h>
+
+class InputQuery : public IQuery
```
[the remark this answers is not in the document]
wu-haoze: Sure. Let me add it in the header files.

deps/CVC4/context/cdhashmap.h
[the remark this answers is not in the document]
MatthewDaggitt: Ah I see, you're defining a convenience method for when we're sure the key should be in there. Yes, defining a subclass would be  much more maintainable way of doing it :+1:

src/input_parsers/PropertyParser.h
wu-haoze: Updated documentation to IQuery

MatthewDaggitt (approved):

Issue #754, opened 2024-02-16
C++ ONNX parser performance issues


MatthewDaggitt:
Follow up from #744 (see the example network).

The C++ parser is currently slower than the ONNX one because my crude implementation of multi-dimensional tensors is not as elegant as those in numpy. In particular the two offending operations are `transpose` and

https://github.com/NeuralNetworkVerification/Marabou/blob/571748e3058b792c8c2eaee1c7381ad8bc878e2d/src/input_parsers/TensorUtils.h#L80-L89

and the all the lookups and broadcasting done here:

https://github.com/NeuralNetworkVerification/Marabou/blob/571748e3058b792c8c2eaee1c7381ad8bc878e2d/src/input_parsers/OnnxParser.cpp#L1637-L1646

I'm a bit torn on how to address this. I can either continue to refine the tensor representation to incorporate strides which allows us to perform these operations much cheaper, or start to depend on an existing tensor library. For the latter case I'm not quite sure what I would use...

wu-haoze:
Hi @MatthewDaggitt , I think we do want to fix this performance issue. If on the small network I sent, parsing is already taking 20+ seconds it's not a good sign. I would be in favor of depending on an existing Tensor library. Do you think it'd be helpful to go to Pytorch or Tensorflow? While this might be a overkill for parsing, it would lay a good foundation for our future agenda of incorporating gradient-based method.

MatthewDaggitt:
I'm not sure whether Pytorch or Tensorflow would help. We need efficient access to array operations over arbitrary-dimension C++ arrays, but I suspect the overhead of setting up an entire Pytorch/Tensorflow neural network just to compute them won't be worth it. Maybe I've misunderstood their capabilities though?

One thing I have found is that in C++ 2023 `mdspan` (multi-dimensional span) was introduced which looks promising:

- https://en.cppreference.com/w/cpp/container/mdspan
- https://www.ashermancinelli.com/std-mdspan-tensors

wu-haoze:
I was suggesting using Pytoch/Tensorflow just for Tensor computation. From pytorch's documentation, they support general operations over tensors independent of machine learning. Though I haven't looked closely at it..

https://pytorch.org/tutorials/beginner/basics/tensorqs_tutorial.html

MatthewDaggitt:
[...]

PR #848, opened 2024-11-12
Attack

mayaswissa:


MatthewDaggitt:
Does this solve the start up time issues described in https://github.com/NeuralNetworkVerification/Marabou/issues/754 when linking libtorch? If so, how did you do it?

idan0610:
> Does this solve the start up time issues described in #754 when linking libtorch? If so, how did you do it?

We are not actually solve this issue, and that's why we added a new CMake flag to allow building marabou without importing the torch library, if someone is not interested in this delay when running multiple number of short queries in a row. But if someone else wants to run a more difficult query, that is expected to take a much longer time, this delay becomes negligible, and the adversarial attack is expected to significantly improve runtimes at least for SAT queriesץ

MatthewDaggitt:
> We are not actually solve this issue, and that's why we added a new CMake flag to allow building Marabou without importing the torch library, if someone is not interested in this delay when running multiple number of short queries in a row. But if someone else wants to run a more difficult query, that is expected to take a much longer time, this delay becomes negligible, and the adversarial attack is expected to significantly improve runtimes at least for SAT queries.

Thanks for you answer @idan0610. While I understand the pragmatics of this decision, this proliferation of build flags seems like a big mistake to me in the long term... What happens if the performance incompatibilities with the PyTorch build options get worse? How do we know how this flag interacts with all the other possible flags for Marabou?

If this is the approach taken, then given that we have known compatibility issues with this build config, my advice would be that at the very least, we should run the whole test suite twice, once with the PyTorch build flag on and once with it off.

MatthewDaggitt:
This should also be documented in the Changelog.

src/engine/Marabou.cpp
```diff
@@ -219,8 +222,13 @@ void Marabou::solveQuery()
 
     struct timespec start = TimeUtils::sampleMicro();
     unsigned timeoutInSeconds = Options::get()->getInt( Options::TIMEOUT );
+
     if ( _engine->processInputQuery( _inputQuery ) )
     {
+        if (_engine->getExitCode() == Engine::ATTACK_SAT)
```
idan0610, line 228: fix coding style

src/engine/Marabou.cpp
```diff
@@ -300,6 +308,11 @@ void Marabou::displayResults( unsigned long long microSecondsElapsed ) const
             printf( "\n" );
         }
     }
+    else if (result == Engine::ATTACK_SAT)
```
idan0610, line 311: same

idan0610:
> Changelog

idan0610:
> > We are not actually solve this issue, and that's why we added a new CMake flag to allow building Marabou without importing the torch library, if someone is not interested in this delay when running multiple number of short queries in a row. But if someone else wants to run a more difficult query, that is expected to take a much longer time, this delay becomes negligible, and the adversarial attack is expected to significantly improve runtimes at least for SAT queries.
> 
> Thanks for you answer @idan0610. While I understand the pragmatics of this decision, this proliferation of build flags seems like a mistake to me in the long term... What happens if the performance incompatibilities with the PyTorch build options get worse? How do we know how this flag interacts with all the other possible flags for Marabou?
> 
> If this is the approach taken then, given that we have known compatibility issues with this build config, my advice would be that at the very least we should run the whole test suite twice - once with the PyTorch build flag on and once with it off.

I understand your concerns, and I'm not sure what is the preferred way to handle this either. @guykatzz what do you think?

guykatzz:
Hi all, and apologies for the delay in my reply.

Matthew, I'm not sure I understand your particular concern about PyTorch. Marabou uses a bunch of other external packages - most notably Gurobi, but also blas, and boost, and possibly a few others. Sure, it is good to run tests with this flag set to both ON and OFF; but are you especially concerned about this package?

MatthewDaggitt:
> Matthew, I'm not sure I understand your particular concern about PyTorch. Marabou uses a bunch of other external packages - most notably Gurobi, but also blas, and boost, and possibly a few others. Sure, it is good to run tests with this flag set to both ON and OFF; but are you especially concerned about this package?

I'm concerned about this package in particular because when we build with Pytorch, we _know_  that the performance of Marabou degrades noticeably (wish we knew why). At least as far as I'm aware there's not a known incompatibility issue with the rest of the code base when building with Gurobi/blas/boost.

MatthewDaggitt:
I'm definitely not in anyway trying to block this PR from merging, but I did want to highlight that it's potentially a hit to maintainability of the code base!

guykatzz:
Can you elaborate on what you mean by "we know that the performance of Marabou degrades noticeably"?

My understanding was that loading PyTorch incurred a certain overhead for each query, but that would only happen if the flag was set to true, right? And in Maya and Idan's experiments we see a significant speedup for certain queries.

MatthewDaggitt:
> Can you elaborate on what you mean by "we know that the performance of Marabou degrades noticeably"?

> My understanding was that loading PyTorch incurred a certain overhead for each query, but that would only happen if the flag was set to true, right?

Yes, this overhead was what I was referring to. Sorry, I should have said "degrades noticeably on some types of queries". 

For context, the added overhead for each query when building with PyTorch can't be a "we're running extra code" issue, but instead is some incompatibility with the build settings of PyTorch and Marabou and should really be fixable.

If we could fix it then we should always build with PyTorch because then we could use the tensor library in the neural network parsers to fix the [existing performance issues](https://github.com/NeuralNetworkVerification/Marabou/issues/754). I could then actually make progress in replacing most of the Python parsers with calls to the C++ parser. However, it was (rightly!) judged that the effects building with the PyTorch flag on was unacceptable.

PR #856, opened 2025-01-19
Fixed deprecated constructors with templates (#855)

### LINKED ISSUE #855: The project cannot be built in non-release mode with >=C++20 and GCC
In your project, you use deprecated template-ids within constructors:
```
.../_deps/marabou-src/src/common/Vector.h:36:14:
warning: template-id not allowed for constructor in C++20 [-Wtemplate-id-cdtor]
   36 |     Vector<T>()
```
Also, your `CMakeLists.txt` defines for non-Clang non-release mode the following:
```
set(COMPILE_FLAGS  -Wall -Wextra -Werror -MMD ) #-Wno-deprecated
```
This causes the warnings to turn into errors.

Since I used your project as a submodule, I have not figured out how to force a different C++ standard just for your project.
What worked out for me was manually modifying the `CMakeLists.txt`.

I do not know what is the situation with Clang.

I would suggest either fixing these constructors, using `-Wno-template-id-cdtor` for GCC (which may not fix it for other compilers), or simply dropping `-Werror`.

Tomaqa:
This PR removes templates appearing within classes that are either useless or, in the case of constructors, forbidden since C++20.

Resolves #855.

### DIFF
--- a/src/common/HashSet.h
+++ b/src/common/HashSet.h
@@ -30,11 +30,11 @@ public:
     typedef typename Super::iterator iterator;
     typedef typename Super::const_iterator const_iterator;
 
-    HashSet<Value>()
+    HashSet()
     {
     }
 
-    HashSet<Value>( const std::initializer_list<Value> &initializerList )
+    HashSet( const std::initializer_list<Value> &initializerList )
         : _container( initializerList )
     {
     }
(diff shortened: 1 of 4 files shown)


Tomaqa:
I rebased the branch in the hope that the recent commits have fixed the previous build errors.

PR #857, opened 2025-01-24
Try to fix dangling reference flagged by CI

MatthewDaggitt:
The CI in various places is failing:
e.g. https://github.com/NeuralNetworkVerification/Marabou/actions/runs/12862710327/job/35858064473#step:12:122

because of a dangling reference. `AttributeProto` object may be deassigned as it is a local, so therefore better simply to return a copy of the `TensorProto` object.

### DIFF
--- a/src/input_parsers/OnnxParser.cpp
+++ b/src/input_parsers/OnnxParser.cpp
@@ -353,7 +353,7 @@ int getRequiredIntAttribute( onnx::NodeProto &node, String name )
     return attr->i();
 }
 
-const onnx::TensorProto getTensorAttribute( onnx::NodeProto &node, String name )
+const onnx::AttributeProto *getTensorAttribute( onnx::NodeProto &node, String name )
 {
     const onnx::AttributeProto *attr =
         findAttribute( node, name, onnx::AttributeProto_AttributeType_TENSOR );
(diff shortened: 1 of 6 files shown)


MatthewDaggitt:
Well even though the tests are still failing, they are failing at a later point so this seems like an improvement?

wu-haoze (approved):
Thanks for fixing this!!

Update: Whoops, just realized it didn't fix the entire CI..

MatthewDaggitt:
No it doesn't fix it completely. I'm unsure why suddenly the compiler is being much more strict? If I get some spare time I'll look at it, but I think it's unlikely to happen in the next couple of weeks...

src/input_parsers/VnnLibParser.cpp
```diff
@@ -20,7 +20,7 @@
 #include "File.h"
 #include "InputParserError.h"
 
-#include <regex>
+#include <boost/regex.hpp>
```
wu-haoze, line 23: Turns out I could reproduce the CI error on my end with the debug build. 

It seems that std::regex is not very stable https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61582. I tried to bump our C++ -std flag but that creates more issues. 

On the other hand using boost/regex seems to fix the issue on my end...

src/input_parsers/OnnxParser.cpp
```diff
@@ -353,15 +353,15 @@ int getRequiredIntAttribute( onnx::NodeProto &node, String name )
     return attr->i();
 }
 
-const onnx::TensorProto getTensorAttribute( onnx::NodeProto &node, String name )
+const onnx::AttributeProto *getTensorAttribute( onnx::NodeProto &node, String name )
```
wu-haoze, line 356: Turns out making copy of TensorProto can be inefficient and caused the CI to time out.

MatthewDaggitt:
Nice finds. Seems to be passing so merge?

PR #860, opened 2025-02-05
Fix a memory leak by introducing destructor for InputQueryBuilder

R-Peleg:
Found by LeakSanitizer. Class members were dynamically allocated but never released.

### DIFF
--- a/src/input_parsers/InputQueryBuilder.cpp
+++ b/src/input_parsers/InputQueryBuilder.cpp
@@ -231,3 +231,42 @@ Equation *InputQueryBuilder::findEquationWithOutputVariable( Variable variable )
     }
     return NULL;
 }
+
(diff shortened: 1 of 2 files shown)


MatthewDaggitt (approved):
Thanks for the spot! This was my fault...

R-Peleg:
Seems from the failing test it is not so simple as I thought, since the Query class uses and releases those objects. Maybe we should use `std::shared_ptr` for handling that?

MatthewDaggitt:
Yup that seems like a reasonable plan. Thanks for the update.

MatthewDaggitt:
Can we try bringing this up to date with `master`. I hope 81d71a2f9fb4b589961d0511ea1472e47a2c8a73 should fix the CI problems.

R-Peleg:
Rebased to current master, including 81d71a2

PR #861, opened 2025-02-06
Backward Analysis using Preimage Approximation

ido-shm-uel:
* Implemented PreimageApproximation Algorithm (simplified, without branch-and-bound) for ReLU, LeakyReLU, Sign and Billinear Layers (arXiv:2305.03686v4 [cs.SE], arXiv:2405.21063 [cs.LG]).
* Added option to perform LP relaxation using backward-preimage-approx.
* Added unit tests at Test_NetworkLevelReasoner.h.


src/engine/PolygonalTightening.h
```diff
@@ -0,0 +1,101 @@
...
+    /*
+      Whether the tightening tightens the
+      lower bound or the upper bound.
+    */
+    PolygonalBoundType _type;
+
+    /*
+      Equality operator.
+    */
+    bool operator==( const PolygonalTightening &other ) const
```
wu-haoze, line 60: Could you point to me where this equality operator is used?

src/engine/PolygonalTightening.h
```diff
@@ -0,0 +1,101 @@
...
+ ** Top contributors (to current version):
+ **   Duligur Ibeling, Guy Katz, Ido Shmuel
+ ** This file is part of the Marabou project.
+ ** Copyright (c) 2017-2024 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved. See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** [[ Add lengthier description here ]]
+
```
wu-haoze, line 13: Could you add a description of what this class is supposed to represent? Does this represent a linear symbolic bound in terms of neurons? Also, this file should probably be in `src/nlr/` since it's very NLR specific.

wu-haoze (changes requested):
Hi @ido-shm-uel , thanks for this addition. I have some questions and (mostly minor) code style suggestions. Please see my comments!

wu-haoze (approved):

PR #889, opened 2025-09-10
Towards Proof Minimization

omriisack:
Renaming SMTCore to SearchTreeHandler; add information stored in lemmas; add function to compute explanation lower bound;


wu-haoze (approved):
LGTM. Could you add to the change log the updates that you made?

PR #890, opened 2025-10-15
Proof Minimization

OmriIsacHUJI:
Introduce GroundBoundManager module, and the proof minimization algorithm


src/configuration/GlobalConfiguration.cpp
```diff
@@ -119,8 +119,8 @@ const bool GlobalConfiguration::WRITE_JSON_PROOF = false;
 const unsigned GlobalConfiguration::BACKWARD_BOUND_PROPAGATION_DEPTH = 3;
 const unsigned GlobalConfiguration::MAX_ROUNDS_OF_BACKWARD_ANALYSIS = 10;
 
-const bool GlobalConfiguration::ANALYZE_PROOF_DEPENDENCIES = false;
-const bool GlobalConfiguration::MINIMIZE_PROOF_DEPENDENCIES = false;
+const bool GlobalConfiguration::ANALYZE_PROOF_DEPENDENCIES = true;
+const bool GlobalConfiguration::MINIMIZE_PROOF_DEPENDENCIES = true;
```
guykatzz, line 123: Just making sure this is as intended. I'm guessing proofs are generally off by default, but if they're on - these should be on?

guykatzz (approved):

src/configuration/GlobalConfiguration.cpp
omriisack: Exactly. If proof production is off (default), these configurations are meaningless

